PR #16, opened 2026-06-26
fix(detectors): eliminate deep-scan-audit false positives

skhan75:
A deep scan of a real TypeScript browser-extension repo surfaced several detector false positives. This corrects them across the static, drift, and Code DNA layers, and makes the Fix Plan projection internally consistent. Each fix ships with a regression test and was re-verified against the source repo.

## Fixes
- **codedna** — capture the full function body when the return type contains braces (`: { ... }` / `Promise<{...}>`) instead of truncating to the annotation; stop exact-duplicate over-matching on trivial bodies that differ only by literal values; require a distinctive op-sequence before reporting a near-exact match.
- **dead-code** — count same-file usage and trim single-name barrel re-exports before flagging an export unused; only count code as unreachable after a complete (`;`-terminated) return/throw, so multi-line continuations no longer false-positive.
- **dependencies** — recognize usage in build config (e.g. webpack) before reporting a dependency as phantom.
- **return-shape** — classify a file by its dominant on-error shape and ignore defensive rethrows; require an explicit error field for error-object returns so plain state objects aren't mislabeled.
- **naming** — treat single-word filenames as convention-neutral.
- **logging** — never recommend a logging library that isn't in the project.
- **fix plan** — present the projection in a single scale (projected score + delta); label not-measured categories honestly.
- **scan --json** — emit only valid JSON on stdout (info messages go to stderr).

## Validation
Re-verified on the source repo: the previously-false findings are eliminated and genuine findings (real cross-file duplicates, truly-dead exports, security smells) are preserved — no recall loss.


src/analyzers/dead-code.ts
```diff
@@ -22,7 +22,7 @@
 
 import type { Analyzer } from "./base.js";
 import type { AnalysisContext, Finding, SourceFile } from "../core/types.js";
-import { buildImportGraph, type FileExport } from "../core/import-graph.js";
+import { buildImportGraph, fileBasename, sourceLookupKey, type FileExport } from "../core/import-graph.js";
 
 const ENTRY_POINT_BASES = new Set([
   "index", "main", "app", "server", "mod", "lib", "init", "__init__",
```
skhan75, line 28: we should have different dead code analyzers per language, and not jumble them into one class

src/analyzers/dead-code.ts
```diff
@@ -42,6 +53,28 @@ function isEntryFile(filePath: string): boolean {
...
 }
 
+/**
+ * Files referenced as a runtime entry point — loaded as a Web Worker or via a
+ * build-output URL — have no static importers by design and must not be flagged
+ * as orphans. We scan every source for `new Worker('X')` / `getURL('X')` string
+ * literals, reduce each to a basename lookup key, and return the set of such
+ * keys. A file whose basename matches is treated as a root.
+ */
+function collectWorkerRootBases(files: SourceFile[]): Set<string> {
```
skhan75, line 63: I guess we are doing this to explicitly flag file paths as imports, so import graphs would not see them as orphans. If that's the case, shouldn't we just create a local top level graph, with import graph as sub graph. And then collect all orphaned nodes with no edges  as dead code nodes? This is just a high level idea

src/analyzers/dependencies.ts
```diff
@@ -20,6 +20,15 @@ const JS_IMPORT_PATTERNS = [
 // Test fixture directories — their imports are intentionally broken/nonexistent
 const FIXTURE_PATH_PATTERN = /(?:fixtures?|testdata|__fixtures__|__mocks__)[/\\]/i;
 
+// Build/config files reference dependencies in ways the import patterns miss:
+// require.resolve('buffer/'), loader: 'ts-loader', plugin string names, etc.
+// Match common bundler/tooling config filenames anywhere in the tree.
+const BUILD_CONFIG_PATTERN =
```
skhan75, line 26: same! Couldn't we have created a dependency graph instead?

src/analyzers/dead-code.ts
skhan75: Agreed. It's one `Analyzer` registration today, but the language paths are already separate functions — `analyzeJsGraph` (JS/TS, real import-graph based), `analyzeGoDeadExports`, and `analyzePythonDeadCode` (the last two still on the weaker substring method). Splitting them into per-language modules behind a thin dispatcher is a clean, low-risk refactor — and the natural place to upgrade Go/Python off substring matching to real import resolution.

Kept it out of this PR since the scope here was eliminating the false positives, not restructuring. Parked the split (+ the Go/Python upgrade) in the backlog.



— 🤖 posted by [Claude Code](https://claude.com/claude-code) on @skhan75's behalf

src/analyzers/dead-code.ts
skhan75: Yes, that's exactly the intent — and your graph framing is the better design. The import graph only models `import` edges today; the worker refs (`new Worker('x')` / `getURL('x')`) are handled as a side-channel set of "root basenames" that we then subtract from the orphan check.

Making them first-class edges — a runtime-load edge from the referencing file to the worker file — would make "dead code = nodes with no incoming edges" fall out for free, and it'd also get transitive reachability right (a worker that's only loaded by another orphan should still be flagged).

One wrinkle to design for: the worker string is a build-output path, so the edge needs resolving that path back to the source file. Basename matching is the quick heuristic we use now; proper resolution (like the import resolver) would be more robust. I'll fold worker-refs + entry-point roots into the import graph as edges in the follow-up instead of this side-channel.



— 🤖 posted by [Claude Code](https://claude.com/claude-code) on @skhan75's behalf

src/analyzers/dependencies.ts
skhan75: Same direction, agreed. `dependencies` is set-difference today (declared in package.json − imported names − build-config refs), not a graph — and it overlaps a lot with dead-code, since both are fundamentally "what's declared vs what's actually reachable."

The clean version is one shared usage/reachability graph: nodes = source files + external packages, edges = imports + build-config refs + dynamic/worker loads. Then phantom-deps and dead-files are both just queries over it (a package node with no inbound usage; a file node with no inbound edge). I'd unify dead-code + dependencies onto that graph in the follow-up.

For this PR I just taught the existing checker to credit build-config usage (e.g. `require.resolve`, loader/plugin strings) so the immediate false positives stop.



— 🤖 posted by [Claude Code](https://claude.com/claude-code) on @skhan75's behalf

PR #20, opened 2026-06-26
fix(cli): exit promptly after a scan (drop the 10-minute report server)

skhan75:
### The problem
`vibedrift <dir>` felt like it took minutes even though the scan finishes in ~1s. The html report path started a localhost server and **kept the process alive for 600 seconds** (`setTimeout(..., 600_000)`) before returning, so the command never came back to the prompt promptly.

### The fix
Write the report files, open the report best-effort (interactive only, detached, skipped in CI / with `VIBEDRIFT_NO_BROWSER=1`), print the path, and exit. The report is self-contained and links to the detailed file by relative path, so `file://` works without a server.

### Measured
- Scan analysis: ~1.3s (within the "< 2s" claim).
- Large-repo authenticated command: **~5s and exits** (scan + HTML render + dashboard upload + startup), was 10 minutes.
- 684 tests pass.

### DIFF
--- a/src/cli/commands/scan.ts
+++ b/src/cli/commands/scan.ts
@@ -747,28 +747,25 @@ async function renderToFormat(
     await writeFile(outputPath, summaryHtml);
     await writeFile(detailedPath, detailedHtml);
     // Serve on localhost
-    const { createServer } = await import("http");
-    const server = createServer((req, res) => {
-      res.writeHead(200, { "Content-Type": "text/html; charset=utf-8" });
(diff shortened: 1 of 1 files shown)


skhan75:
Superseded by #21. That PR removes the 10-minute report server AND links authenticated scans to the dashboard instead of a local HTML report (this file-open approach is kept as the offline fallback there).

— 🤖 closed by [Claude Code](https://claude.com/claude-code) on @skhan75's behalf

PR #27, opened 2026-07-06
fix(core): normalize relativePath to forward slashes on Windows

### LINKED ISSUE #26: [bug] Import graph produces zero incoming-counts on Windows (backslash path separator)
## What happened

Running VibeDrift on Windows reports 146 orphan files and 128 phantom exports — virtually every file in the project. These are false positives caused by a path separator mismatch in the import graph.

`fileBasename()` in `src/core/import-graph.ts` splits on `/` only:

```typescript
export function fileBasename(filePath: string): string {
  const file = filePath.split("/").pop() ?? filePath;
```

But `relativePath` is assigned via Node's `path.relative()` in `src/core/discovery.ts`, which on Windows produces backslash-separated paths (e.g., `src\analyzers\complexity.ts`). When `split("/")` is called on a backslash path, it returns the entire string as one element, so `fileBasename` returns the full path instead of just the filename.

This means `sourceLookupKey("./complexity.js")` returns `"complexity"`, but `fileBasename("src\\analyzers\\complexity.ts")` returns `"src\\analyzers\\complexity.ts"`. They never match, so every file gets `incomingCount === 0`.

Affected detectors:
- `dead-code` analyzer — orphan file detection and unused export detection both rely on `buildImportGraph`
- `drift-phantom_scaffolding` — also uses `buildImportGraph` → `incomingCount`

## Steps to reproduce

1. On Windows, clone the VibeDrift repo
2. Run `npm ci && npm run build`
3. Run `node bin/vibedrift.mjs --json --local-only`
4. Observe the dead-code finding: "146 file(s) with zero incoming imports"
5. Observe the phantom-scaffolding finding: "128 phantom export(s)"
6. Verify any flagged function is actually imported (e.g., `createAnalyzerRegistry` is imported in `src/cli/commands/scan.ts`, `test/calibration/precision-recall.ts`, `test/calibration/run.ts`, and `test/integration/scan.test.ts`)

## Expected behavior

Files imported through relative paths (e.g., `import { complexityAnalyzer } from "./complexity.js"`) should have their `incomingCount` incremented correctly regardless of OS. The 146 orphan-file and 128 phantom-export findings should not appear when all files are reachable.

## Environment

- VibeDrift version: 0.14.8
- Node version: v24.12.0
- OS: Windows 11

## Scan output or logs

```
[error] drift-phantom_scaffolding: DRIFT: ./: 128 phantom export(s) — CRUD-named functions never imported and never routed
[warning] dead-code: 146 file(s) with zero incoming imports (potential dead code)
[error] dead-code: 79 exported symbols appear unused: configExists, runCodeDnaAnalysis, loadVibeDriftConfig...
```

Example false positive — `createAnalyzerRegistry()` flagged as phantom export with "no incoming imports, not in any route table", but grep confirms it's imported in 4 files:

```
src/cli/commands/scan.ts:8:import { createAnalyzerRegistry } from "../../analyzers/index.js";
test/calibration/precision-recall.ts:19:import { createAnalyzerRegistry } from "../../src/analyzers/index.js";
test/calibration/run.ts:16:import { createAnalyzerRegistry } from "../../src/analyzers/index.js";
test/integration/scan.test.ts:5:import { createAnalyzerRegistry } from "../../src/analyzers/index.js";
```

## Additional context

The root cause is in `src/core/discovery.ts` line 87:

```typescript
const relPath = relative(rootDir, fullPath);
```

Node's `path.relative()` returns OS-native separators. A minimal fix would be normalizing at the source:

```typescript
const relPath = relative(rootDir, fullPath).replace(/\\/g, "/");
```

Alternatively, `fileBasename()` and `sourceLookupKey()` in `src/core/import-graph.ts` could split on both separators:

```typescript
const file = filePath.split(/[/\\]/).pop() ?? filePath;
```

The `directoryOf()` helper in `src/drift/utils.ts` has the same forward-slash-only assumption and may also be affected on Windows.

BrandonGarate177:
## Summary

On Windows, `path.relative()` returns backslash-separated paths (e.g., `src\analyzers\complexity.ts`). The import graph's `fileBasename()` splits on `/` only, so on Windows it returns the full path instead of the filename. This causes `incomingCount` to be 0 for every file, producing ~270 false positives from the dead-code and phantom-scaffolding detectors.

**Root cause:** `src/core/discovery.ts` assigns `relativePath` via `path.relative()` without normalizing separators. Downstream code (import-graph, dead-code, phantom-scaffolding, MCP tools) assumes forward slashes.

**Fix:** Normalize `relativePath` to forward slashes at the single point of creation in `discoverFiles()`, apply the same normalization in the three tool commands that compute their own relative paths at runtime, and harden `fileBasename()` to split on both `/` and `\` as defense-in-depth.

**Before (Windows):**
- 146 orphan files (false positives — every file flagged)
- 128 phantom exports in one finding (false positives — every export flagged)
- 79 unused exports (false positives)

**After (Windows):**
- 11 orphan files (legitimate: entry points, scripts, dynamic imports)
- ~66 phantom exports across 9 directory-scoped findings (legitimate: mostly test fixture functions that are intentionally never imported)
- Vibe Drift Score: 91.3 → 91

Also fixes git metadata grafting, which was comparing backslash `relativePath` against git's forward-slash output — meaning temporal weighting and pivot detection were silently broken on Windows.

### Files changed

| File | Change |
|------|--------|
| `src/core/discovery.ts` | `.replace(/\\/g, "/")` on `relativePath` assignment |
| `src/core/import-graph.ts` | `fileBasename()` splits on `/[/\\]/` (defense-in-depth) |
| `src/tools-core/tools/check-file-drift.ts` | Same normalization on runtime `relative()` call |
| `src/tools-core/tools/validate-change.ts` | Same |
| `src/cli/commands/watch.ts` | Same |
| `test/unit/core/import-graph.test.ts` | 4 regression tests: multi-component paths + explicit backslash cases |

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Refactor
- [ ] Test
- [ ] Chore / infra

## Checklist

## Related issues

Closes #26

### DIFF
--- a/src/core/discovery.ts
+++ b/src/core/discovery.ts
@@ -85,7 +85,7 @@ export async function discoverFiles(rootDir: string): Promise<{ files: SourceFil
       }
 
       const fullPath = join(dir, entry.name);
-      const relPath = relative(rootDir, fullPath);
+      const relPath = relative(rootDir, fullPath).replace(/\\/g, "/");
 
       if (entry.isDirectory()) {
         if (SKIP_DIRS.has(entry.name) || entry.name.startsWith(".")) continue;
--- a/src/cli/commands/watch.ts
+++ b/src/cli/commands/watch.ts
@@ -295,7 +295,7 @@ async function snapshotMtimes(rootDir: string): Promise<Map<string, number>> {
     }
     for (const entry of entries) {
       const full = join(dir, entry);
-      const rel = relative(rootDir, full);
+      const rel = relative(rootDir, full).replace(/\\/g, "/");
       if (shouldIgnore(rel)) continue;
       let info;
       try {
--- a/src/core/import-graph.ts
+++ b/src/core/import-graph.ts
@@ -163,10 +163,10 @@ export function sourceLookupKey(source: string): string {
 
 /** Lookup key derived from a file's path, used to match against import sources. */
 export function fileBasename(filePath: string): string {
-  const file = filePath.split("/").pop() ?? filePath;
+  const file = filePath.split(/[/\\]/).pop() ?? filePath;
   const noExt = file.replace(/\.(?:ts|tsx|js|jsx|mjs|cjs)$/, "");
   if (noExt === "index") {
-    const parts = filePath.split("/");
+    const parts = filePath.split(/[/\\]/);
     return parts.length >= 2 ? parts[parts.length - 2] : noExt;
   }
   return noExt;
--- a/src/drift/architectural-contradiction.ts
+++ b/src/drift/architectural-contradiction.ts
@@ -51,7 +51,7 @@ function detectDataAccess(file: DriftFile): { pattern: DataAccessPattern; eviden
 
   // Raw SQL
   const sqlEvidence = extractEvidence(c, /(?:SELECT|INSERT|UPDATE|DELETE)\s+(?:FROM|INTO|SET|\*)\b/gi);
-  if (sqlEvidence.length > 0 && !isRepoFile(file.path)) {
+  if (sqlEvidence.length > 0 && !isRepoFile(file.relativePath)) {
     results.push({ pattern: "raw_sql", evidence: sqlEvidence });
   }
 
@@ -62,7 +62,7 @@ function detectDataAccess(file: DriftFile): { pattern: DataAccessPattern; eviden
 
   // Direct DB
   const dbEvidence = extractEvidence(c, /\b(?:db|pool|client)\.\s*(?:Query|Exec|QueryRow|query|execute|raw)\s*\(/g);
-  if (dbEvidence.length > 0 && !isRepoFile(file.path)) {
+  if (dbEvidence.length > 0 && !isRepoFile(file.relativePath)) {
     results.push({ pattern: "direct_db", evidence: dbEvidence });
   }
 
@@ -157,7 +157,7 @@ function detectDIPattern(file: DriftFile): { pattern: DIPattern; evidence: Evide
 
 function buildProfile(file: DriftFile): FileArchProfile | null {
   if (!file.language) return null;
-  if (!isAnalyzableSource(file.path)) return null;
+  if (!isAnalyzableSource(file.relativePath)) return null;
 
   const dataAccess = detectDataAccess(file);
   const errorHandling = detectErrorHandling(file);
@@ -167,7 +167,7 @@ function buildProfile(file: DriftFile): FileArchProfile | null {
   // Only include files that have meaningful patterns
   if (dataAccess.length === 0 && errorHandling.length === 0 && config.length === 0) return null;
 
-  return { file: file.path, language: file.language, dataAccess, errorHandling, config, di };
+  return { file: file.relativePath, language: file.language, dataAccess, errorHandling, config, di };
 }
 
 /**
@@ -225,7 +225,7 @@ const DATA_ACCESS_PRIORITY: DataAccessPattern[] = ["raw_sql", "direct_db", "http
  * stays in lockstep with this detector's vocabulary.
  */
 export function classifyDataAccessLabel(content: string, path: string): string | null {
-  const hits = detectDataAccess({ content, path, language: "typescript" } as DriftFile);
+  const hits = detectDataAccess({ content, relativePath: path, language: "typescript" } as DriftFile);
   if (hits.length === 0) return null;
   const primary = hits
     .slice()
--- a/src/drift/async-consistency.ts
+++ b/src/drift/async-consistency.ts
@@ -30,7 +30,7 @@ interface FileAsyncProfile {
 
 function analyzeAsync(file: DriftFile): FileAsyncProfile | null {
   if (!file.language || !["javascript", "typescript"].includes(file.language)) return null;
-  if (!isAnalyzableSource(file.path)) return null;
+  if (!isAnalyzableSource(file.relativePath)) return null;
 
   // Counting + classification live in the shared async-style module so the
   // detector and the MCP validate_change tool agree on the vocabulary. Evidence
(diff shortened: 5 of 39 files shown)


skhan75:
@anishekkamal  can you test this fix on your windows?

src/core/discovery.ts
```diff
@@ -85,7 +85,7 @@ export async function discoverFiles(rootDir: string): Promise<{ files: SourceFil
       }
 
       const fullPath = join(dir, entry.name);
-      const relPath = relative(rootDir, fullPath);
+      const relPath = relative(rootDir, fullPath).replace(/\\/g, "/");
```
skhan75, line 88: This fixes relativePath, but the sibling path field just below (line 111, path: fullPath) is left un-normalized, and phantom-scaffolding.ts:102 builds its import graph from f.path, not relativePath. So the phantom detector still sees backslashes on Windows and stays broken. The "128 → 1 phantom" claim can't come from this diff. Either also normalize path here, switch phantom to f.relativePath, or normalize inside fileBasename/sourceLookupKey so both fields are covered.

skhan75 (changes requested):
Nice catch on the git-metadata and dead-code side. Before this merges, I don't think the phantom-scaffolding part of the fix actually lands, and the "128 → 1 phantom exports" number needs a second look.

The phantom detector doesn't read relativePath. In src/drift/phantom-scaffolding.ts:102, it builds its import graph from f.path:

```
.map((f) => ({
  path: f.path,
  relativePath: f.path, // uses f.path, not f.relativePath
  ...
}))
```

This PR normalizes relativePath at discovery.ts:88, but never touches f.path (set at discovery.ts:111 as path: fullPath, which stays backslashes on Windows). So on Windows, the phantom detector still feeds backslash paths into fileBasename, incomingCount stays 0 for every file, and the phantom false positives persist. I confirmed the mechanism: fileBasename("src\\analyzers\\complexity.ts") returns the whole string instead of "complexity", and incomingCount comes back 0 for a file that is imported.

Since phantom's input doesn't change in this PR, the "128 → 1" result can't be coming from these 5 files. Could you re-verify that number on a Windows machine, or drop the phantom claim from the description?

Two Ways to Make It Real:
Minimal: change line 102 to relativePath: f.relativePath so phantom uses the already-normalized field.
Sturdier: make sourceLookupKey/fileBasename in import-graph.ts split on both separators (/[\\/]/). That fixes dead-code, phantom, and any future caller in one place. Keep the discovery.ts normalization too, since git-metadata still needs the string match.
Separately, the two added tests use forward-slash paths, so they pass with or without this fix. A real regression test should feed a backslash path.

BrandonGarate177:
I re-ran the scan on both commits and the "128 → 1" in the original description was not true. I corrected it:

**Before:** 1 phantom finding reporting 128 phantom exports — all false positives because `incomingCount` was 0 for every file.

**After:** 9 phantom findings reporting ~66 phantom exports scoped per directory — these are legitimate (mostly test fixture functions that are intentionally never imported). The false positives are eliminated; real findings now correctly surface.

Updated the PR description to reflect this.

On the phantom data flow, im convinced the fix does reach phantom-scaffolding:

1. `discovery.ts` normalizes `SourceFile.relativePath` (this PR)
2. `buildDriftContext()` in `drift/index.ts:44` sets `DriftFile.path = f.relativePath` (the normalized value)
3. `phantom-scaffolding.ts:102` sets `relativePath: f.path` — this `f.path` is `DriftFile.path`, which IS the normalized relative path, not the absolute `SourceFile.path`

The naming is confusing (`DriftFile.path` ≠ `SourceFile.path`), but the normalized value flows through. Confirmed by running the scan on both commits.

Anyways, I took your suggestion and hardened `fileBasename()` to split on `/[/\\]/` as defense-in-depth. Also added explicit backslash regression tests that feed `"src\\utils\\helper.ts"` paths to both `fileBasename` and `buildImportGraph`.

src/core/discovery.ts
BrandonGarate177: The phantom detector's `f.path` at line 102 is `DriftFile.path`, not `SourceFile.path`. It's set from `SourceFile.relativePath` in `drift/index.ts:44`:

```typescript
files: ctx.files.map((f) => ({
  path: f.relativePath,  // ← normalized by this PR
  ...
}))
```

So the normalization does reach phantom-scaffolding through the drift context. The absolute `SourceFile.path` (set at discovery.ts:111 as `fullPath`) is never used by the phantom detector.

I also hardened `fileBasename()` to split on both separators (`/[/\\]/`) so even if a future caller bypasses the discovery normalization, the import graph handles it. Backslash-specific regression tests added to prove the behavior directly.

skhan75: Ah, you're right, I mixed up the two path fields. Phantom reads DriftFile.path, which is built from relativePath, so your fix does reach it. The `fileBasename` hardening and the new tests cover the rest. One small ask, can we rename `DriftFile.path` to `relativePath` in a follow-up so nobody trips on this again? LGTM.

skhan75 (commented):
LGTM, just one minor nit on the rename, and we are good to go.

**Just Fyi**
Now thinking about this more, we have a dependency graph build on the imports, called import graph, but the weak part is how its edges get built. Imports are parsed with regexes and matched to files by basename, so any two files with the same name share edges. We should move this to tree-sitter parsing with real module resolution. I'll open a detailed issue for it.

src/core/discovery.ts
BrandonGarate177: Okay I just renamed the `DriftFile.path` field to -> `relativePath` across all 17 source files and 15 test files. `DeviatingFile.path` and route URL paths left untouched.

skhan75 (approved):

PR #30, opened 2026-07-06
fix(deps): ignore import-like text in dependency scan

### LINKED ISSUE #28: [bug] Dependency analyzer matches import patterns inside comments and string literals
## What happened

The dependency analyzer reports "12 packages imported but not in package.json" — all false positives. The detected "packages" (`@vibedrift/cli`, `lodash`, `interesting`, `no surface`, `pkg`) come from JSDoc comments, code documentation, and regex pattern strings, not actual import statements.

The `collectImportedPackages()` function in `src/analyzers/dependencies.ts` runs its import-matching regexes against raw `file.content` without stripping comments or string/regex literals first.

## Steps to reproduce

1. Run `node bin/vibedrift.mjs --json --local-only` on the VibeDrift repo itself
2. Observe: "12 packages imported but not in package.json: @vibedrift/cli, pkg, lodash, interesting, no surface..."
3. Inspect the locations:
   - `src/render.ts:5` — JSDoc comment: `import { renderHtmlReport } from "@vibedrift/cli/render"`
   - `src/analyzers/imports.ts:9` — documentation: `` `require('lodash')` (non-builtin npm package) ``
   - `src/output/terminal.ts:142` — prose: `the shift from "interesting" to "I should fix this"`
   - `src/scoring/engine.ts:221` — prose: `"clean" from "no surface"`
   - `src/analyzers/dependencies.ts:16` — regex literal: `/require\(\s*['"]([^./][^'"]*)['"]\s*\)/g`

None of these are actual import statements.

## Expected behavior

Import-matching regexes should only match actual code, not content inside `//` comments, `/* */` block comments, string literals, or regex literals.

## Environment

- VibeDrift version: 0.14.8
- Node version: v24.12.0
- OS: Windows 11 (but this is platform-independent)

## Scan output or logs

```
[error] dependencies: 12 packages imported but not in package.json: @vibedrift/cli, pkg, lodash, interesting, no surface...
```

Locations all point to comments or string literals, not actual import/require statements.

## Additional context

The `codedna/function-extractor.ts` tokenizer already strips comments before processing:

```typescript
cleaned = cleaned.replace(/\/\/[^\n]*/g, "");
cleaned = cleaned.replace(/\/\*[\s\S]*?\*\//g, "");
```

A similar approach in `collectImportedPackages()` would fix this. Strip single-line comments, block comments, and string literals from `file.content` before running `JS_IMPORT_PATTERNS`.

rathodkunj2005:
## Summary

- Mask comments and non-import string/template literal text before applying JS dependency import regexes.
- Keep legitimate static `import`, `require(...)`, and dynamic `import(...)` specifier strings detectable.
- Treat the package's own `package.json` name as declared so self-references like `@vibedrift/cli/tools` are not reported missing.

Fixes #28.

## Verification

- `npm test -- test/unit/analyzers/dependencies.test.ts` → 7 tests passed
- `npm run typecheck` → passed
- `npm run build` → build success
- `node bin/vibedrift.mjs --json --local-only > /tmp/vibedrift-self.json; ...` → `no issue-28 prose/comment false positives`
- `npm run lint` → passed with existing warnings (128 warnings, 0 errors)
- `npm test` → 114 test files passed, 720 tests passed

### DIFF
--- a/src/analyzers/dependencies.ts
+++ b/src/analyzers/dependencies.ts
@@ -1,5 +1,5 @@
 import type { Analyzer } from "./base.js";
-import type { AnalysisContext, Finding } from "../core/types.js";
+import type { AnalysisContext, Finding, SyntaxNode } from "../core/types.js";
 
 const NODE_BUILTINS = new Set([
   "assert", "buffer", "child_process", "cluster", "console", "constants",
@@ -11,10 +11,14 @@ const NODE_BUILTINS = new Set([
 ]);
 
 const JS_IMPORT_PATTERNS = [
-  /(?:import|from)\s+['"]([^./][^'"]*)['"]/g,
-  /require\(\s*['"]([^./][^'"]*)['"]\s*\)/g,
-  // Dynamic imports: await import("pkg") or import("pkg")
-  /import\(\s*['"]([^./][^'"]*)['"]\s*\)/g,
+  // ESM imports/exports, including bare side-effect imports after any line.
+  /^\s*(?:import|export)\s+(?:[^;\n]*?\s+from\s+)?['"]([^./][^'"]*)['"]/gm,
+  // CommonJS require in expression/statement positions. This intentionally
+  // avoids matching regex sources or prose strings that merely mention require().
+  /(?:^|[=(:,{[]\s*|\breturn\s+)require\(\s*['"]([^./][^'"]*)['"]\s*\)/gm,
+  // Dynamic imports: await import("pkg") or import(`pkg`). Template literals
+  // count only when the specifier is a literal package name with no ${...}.
+  /(?:^|[=(:,{[]\s*|\bawait\s+|\breturn\s+)import\(\s*['"`]([^./][^'"`$]*)['"`]\s*\)/gm,
 ];
 
 // Test fixture directories — their imports are intentionally broken/nonexistent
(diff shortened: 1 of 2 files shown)


anishekkamal:
Will look into this.

src/analyzers/dependencies.ts
```diff
@@ -64,6 +64,83 @@ function extractJsPackageName(specifier: string): string {
   return specifier.split("/")[0];
 }
 
+function isImportSpecifierString(prefix: string): boolean {
+  return /(?:\bimport\s*\(|\brequire\s*\(|\bfrom|^\s*import)\s*$/.test(prefix);
```
skhan75: This regex only recognizes a bare import "pkg" when it is the very first thing in the file. Any bare side-effect import that comes after another line gets its package name erased, so it's dropped from the scan. If that package is in package.json it now gets falsely reported as "unused," and if it isn't, the real missing dependency is silently missed.

A "bare import" is one with no variable, used only for side effects. Real, common examples like - import "reflect-metadata", import "dotenv/config", import "core-js/stable".

For example - 

```
import express from "express";
import "reflect-metadata";
```

`reflect-metadata` is a genuine dependency. This code only lets the analyzer see it if it's literally the first line of the file. Because import express comes first, `reflect-metadata` gets wiped out. Result: if reflect-metadata is in package.json, VibeDrift wrongly says "reflect-metadata is unused." If it isn't, VibeDrift stops warning you it's missing. I tested this: the old code sees it, this PR doesn't.

src/analyzers/dependencies.ts
```diff
@@ -64,6 +64,83 @@ function extractJsPackageName(specifier: string): string {
...
+          i += 2;
+          break;
+        }
+        masked += content[i] === "\n" ? "\n" : " ";
+        i++;
+      }
+      continue;
+    }
+
+    if (ch === "'" || ch === '"' || ch === "`") {
```
skhan75: The scanner walks the file character by character. When it sees a ' or ", it assumes "a string just started" and blanks everything until the next matching quote. But quotes also appear inside regular expressions and JSX text, where they are not strings.

Examples -
 
```js
const re = /['"]/;
import fs from "fs-extra";
```

Same thing with an apostrophe in JSX text:

```jsx
const Note = () => <p>don't forget</p>;
const chalk = require("chalk");
```

The apostrophe in "don't" starts a fake string that swallows the require("chalk") below it. This is the worst one because it is silent and unpredictable. A single stray apostrophe anywhere can make a real dependency disappear.

src/analyzers/dependencies.ts
```diff
@@ -64,6 +64,83 @@ function extractJsPackageName(specifier: string): string {
...
+      const quote = ch;
+      const keep = quote !== "`" && isImportSpecifierString(masked);
```
skhan75: This line throws away everything inside backtick template strings. Real imports can live inside backticks too, so they get lost.

skhan75 (changes requested):
Hi @rathodkunj2005 , nice work spotting a real bug here and backing it up with tests and a full verification run. One thing that would level up this PR and your future ones is to check what the repo already ships before hand-writing something like the `maskJsNonCodeImportText` scanner, because we already depend on `tree-sitter` and parse every file into an AST, and pulling imports from that tree would sidestep this whole class of edge cases. Right now, for example, a quote inside a regex or a word like don't in JSX gets mistaken for a string and quietly erases the real imports below it, which a parser would never do. Also worth building as a habit is testing the opposite direction whenever a fix makes something stop matching, since this one ends up dropping genuine imports like import "reflect-metadata" along with the noise.

src/analyzers/dependencies.ts
rathodkunj2005: Fixed in da77424 — the scan now walks the tree-sitter AST (`import_statement`/`export_statement` nodes), so bare side-effect imports like `import "reflect-metadata"` are picked up regardless of position. Added a regression test: "keeps bare side-effect imports after other imports".

src/analyzers/dependencies.ts
rathodkunj2005: Fixed in da77424 — the character-by-character masker is gone entirely. Imports are read from the parsed AST, so quotes inside regex literals and JSX text can't swallow real imports. Regression test: "does not let regex quotes or JSX apostrophes hide later imports".

src/analyzers/dependencies.ts
rathodkunj2005: Fixed in da77424 — template strings are no longer blanked. Dynamic `import(`pkg`)` with a literal specifier (no ${...}) is now resolved from the AST call node. Regression test: "detects literal template dynamic imports without counting docs in templates".

rathodkunj2005:
@skhan75 Thanks for the detailed review — you were right on all three counts. Addressed in da77424:

- Replaced the hand-written `maskJsNonCodeImportText` scanner with import extraction from the tree-sitter AST the repo already parses, as you suggested. The regex patterns are kept only as a fallback for files that fail to parse.
- Bare side-effect imports (`import "reflect-metadata"`) are now detected at any position in the file.
- Quotes inside regex literals and JSX text no longer swallow later imports, and template strings aren't blanked.

Each case has a dedicated regression test (63 new test lines); full suite passes (723 tests) and typecheck is clean. Ready for another look.

rathodkunj2005:
@skhan75 conflicts are resolved and the branch is back up to date with `main`.

What changed in the merge resolution:
- Kept the repo's new shared `parseImportsAst` path from `main` for dependency extraction instead of the old hand-written scanner.
- Preserved the regression coverage from this PR for the cases you called out: bare side-effect imports after other imports, regex/JSX quotes not hiding later imports, and literal template dynamic imports.
- Added literal template-string support to the AST import extraction so `import(`pkg`)` is counted while docs/prose templates are ignored.

Verification on the resolved branch:
- `npm test -- test/unit/analyzers/dependencies.test.ts` -> 9 passed
- `npm run typecheck` -> passed
- `npm run build` -> passed
- `npm test` -> 128 files / 922 tests passed

Ready for another look.

skhan75 (approved):

skhan75:
Nice work! I like the small PRs, easy to review. Small tip for next time: before adding a rare-case handler, ask yourself if it's just being thorough or actually stopping the tool from being wrong. This one prevents a real false alarm (flagging a used package as unused), so it's a keep. We just try not to be a catch-all, and that question keeps it sharp.

Also if you want to continue to contribute and learn more, I'd advice you to join our [Discord](https://discord.gg/5Nzc52MuR) channel, where you can learn from other contributors, ask questions and directly post feedbacks.

PR #31, opened 2026-07-06
fix(dead-code): track dynamic await import() in the import graph

### LINKED ISSUE #29: [bug] Dead-code analyzer does not track dynamic imports (`await import()`)
## What happened

The dead-code analyzer reports "79 exported symbols appear unused" — many are false positives. Symbols like `runCodeDnaAnalysis`, `runMlAnalysis`, `fetchAiSummary`, and `checkForUpdate` are flagged as unused, but they are imported via dynamic `await import()` in `src/cli/commands/scan.ts`.

The import graph's `IMPORT_PATTERNS` in `src/core/import-graph.ts` only matches static import syntax. It does not match `const { X } = await import("./module.js")`, which `scan.ts` uses extensively (34 dynamic imports) for lazy-loading and startup performance.

## Steps to reproduce

1. Run `node bin/vibedrift.mjs --json --local-only` on the VibeDrift repo itself
2. Observe: "79 exported symbols appear unused: runCodeDnaAnalysis, loadVibeDriftConfig..."
3. Verify these are actually used:
   ```
   src/cli/commands/scan.ts:280:  const { runCodeDnaAnalysis } = await import("../../codedna/index.js");
   src/cli/commands/scan.ts:317:  const { runMlAnalysis } = await import("../../ml-client/index.js");
   src/cli/commands/scan.ts:164:  const { getChangedFiles } = await import("../../core/git-metadata.js");
   ```

## Expected behavior

Exports consumed via `const { X } = await import("./path")` should be counted as used in the import graph, preventing false "unused export" findings.

## Environment

- VibeDrift version: 0.14.8
- Node version: v24.12.0
- OS: Windows 11 (but this is platform-independent)

## Scan output or logs

```
[error] dead-code: 79 exported symbols appear unused: configExists, __test_writeContextIfRequested, runCodeDnaAnalysis, loadVibeDriftConfig, resolveDeviationWeights...
```

Cross-referencing with grep shows many of these are imported dynamically in `scan.ts`.

## Additional context

`scan.ts` uses dynamic imports intentionally for startup performance — static imports would force Node to parse all modules even for `--help` or `--version`. This is a standard CLI optimization pattern.

The fix would be adding a pattern to `IMPORT_PATTERNS` in `src/core/import-graph.ts`:

```typescript
/(?:const|let|var)\s*\{([^}]+)\}\s*=\s*await\s+import\(\s*['"]([^'"]+)['"]\s*\)/g
```

This captures the same two groups (names + source path) as the existing patterns, so `parseImports()` would handle it without further changes.

BrandonGarate177:
## Summary

The import graph's `IMPORT_PATTERNS` in `src/core/import-graph.ts` only matched static import syntax (`import { X } from`, `require()`). Exports consumed via `const { X } = await import("./module.js")` were invisible to the graph, causing the dead-code analyzer to flag them as "unused."

`scan.ts` uses 34 dynamic imports for lazy-loading (standard CLI startup optimization). This produced 27 false "unused export" findings — a 34% false positive rate on the unused-exports detection.

**Fix:** Add one regex pattern to `IMPORT_PATTERNS` that captures destructured dynamic imports:

```typescript
/(?:const|let|var)\s*\{([^}]+)\}\s*=\s*await\s+import\(\s*['"]([^'"]+)['"]\s*\)/g
```

It captures the same two groups (names + source path) as existing patterns, so `parseImports()` handles it without further changes.

**Before:** 79 exported symbols flagged as unused  
**After:** 65 exported symbols flagged as unused (14 false positives eliminated in this test run on Windows)

Note: Testing was performed without the Windows path normalization fix (#26) applied, which limits import graph resolution on Windows. On Unix (where CI runs), this fix eliminates 27 false positives since path resolution works fully. The remaining 65 are a mix of true positives and symbols only reachable via patterns not yet covered (e.g., `import()` without destructuring, conditional re-exports).

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Refactor
- [ ] Test
- [ ] Chore / infra

## Checklist

## Related issues

Closes #29

### DIFF
--- a/src/core/import-graph.ts
+++ b/src/core/import-graph.ts
@@ -59,6 +59,10 @@ const IMPORT_PATTERNS = [
   /import\s*\*\s*as\s+(\w+)\s+from\s*['"]([^'"]+)['"]/g,
   /(?:const|let|var)\s*\{([^}]+)\}\s*=\s*require\(\s*['"]([^'"]+)['"]\s*\)/g,
   /(?:const|let|var)\s+(\w+)\s*=\s*require\(\s*['"]([^'"]+)['"]\s*\)/g,
+  // Dynamic imports: const { X } = await import("./module.js")
+  // and namespace form: const ns = await import("./module.js")
+  /(?:const|let|var)\s*\{([^}]+)\}\s*=\s*await\s+import\(\s*['"]([^'"]+)['"]\s*\)/g,
+  /(?:const|let|var)\s+(\w+)\s*=\s*await\s+import\(\s*['"]([^'"]+)['"]\s*\)/g,
 ];
 
 // Re-export edges: a barrel file's `export { X } from './y'` / `export * from
--- a/test/unit/core/import-graph.test.ts
+++ b/test/unit/core/import-graph.test.ts
@@ -67,3 +67,55 @@ describe("parseExports — single-name re-export trimming", () => {
     expect(names).toContain("internalName");
   });
 });
+
+describe("parseImports — dynamic await import()", () => {
+  it("captures destructured names from `const { X } = await import('./mod')`", () => {
+    const { names, sources } = parseImports(
+      makeFile(`const { runAnalysis, buildGraph } = await import("./codedna/index.js");\n`),
+    );
+    expect(names.has("runAnalysis")).toBe(true);
+    expect(names.has("buildGraph")).toBe(true);
+    expect(sources.has("./codedna/index.js")).toBe(true);
+  });
+
+  it("captures dynamic imports with single quotes", () => {
+    const { names, sources } = parseImports(
+      makeFile(`const { checkForUpdate } = await import('../../core/update-check.js');\n`),
+    );
+    expect(names.has("checkForUpdate")).toBe(true);
+    expect(sources.has("../../core/update-check.js")).toBe(true);
+  });
+
+  it("captures let and var destructured dynamic imports", () => {
+    const { names, sources } = parseImports(
+      makeFile(`let { openBrowser } = await import("../auth/browser.js");\n`),
+    );
+    expect(names.has("openBrowser")).toBe(true);
+    expect(sources.has("../auth/browser.js")).toBe(true);
+  });
+
+  it("captures namespace form `const ns = await import('./mod')`", () => {
+    const { names, sources } = parseImports(
+      makeFile(`const utils = await import("./helpers.js");\n`),
+    );
+    expect(names.has("utils")).toBe(true);
+    expect(sources.has("./helpers.js")).toBe(true);
+  });
+
+  it("captures multi-line destructured dynamic imports", () => {
+    const { names, sources } = parseImports(
+      makeFile(`const {\n  assembleBaseline,\n  writeBaseline\n} = await import("../../core/baseline.js");\n`),
+    );
+    expect(names.has("assembleBaseline")).toBe(true);
+    expect(names.has("writeBaseline")).toBe(true);
+    expect(sources.has("../../core/baseline.js")).toBe(true);
+  });
+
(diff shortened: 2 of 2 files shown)


test/unit/core/import-graph.test.ts
```diff
@@ -67,3 +67,30 @@ describe("parseExports — single-name re-export trimming", () => {
     expect(names).toContain("internalName");
   });
 });
+
+describe("parseImports — dynamic await import()", () => {
```
skhan75, line 71: Consider a negative test asserting const ns = await import('./m') is intentionally not captured (documents the limitation from #2), plus test that a multi-line destructure case (const {\n a,\n b\n} = await import(...)) form works as well

src/core/import-graph.ts
```diff
@@ -59,6 +59,8 @@ const IMPORT_PATTERNS = [
   /import\s*\*\s*as\s+(\w+)\s+from\s*['"]([^'"]+)['"]/g,
   /(?:const|let|var)\s*\{([^}]+)\}\s*=\s*require\(\s*['"]([^'"]+)['"]\s*\)/g,
   /(?:const|let|var)\s+(\w+)\s*=\s*require\(\s*['"]([^'"]+)['"]\s*\)/g,
+  // Dynamic imports: const { X } = await import("./module.js")
+  /(?:const|let|var)\s*\{([^}]+)\}\s*=\s*await\s+import\(\s*['"]([^'"]+)['"]\s*\)/g,
```
skhan75, line 64: Good catch @BrandonGarate177 !

This covers the destructured form only. `const ns = await import('./m')` and bare `await import('./m')` register neither names nor source, so orphan-file detection can still false-positive on a file reachable only that way (the require family has a namespace pattern on line 61 for exactly this reason). 

Fine to defer, but either add the namespace pattern or note the limitation in this comment so the boundary is explicit.

This is a rare case, but good to handle them as well.  The static bare form (import './styles.css', import './register-globals') is everywhere. The dynamic bare form is uncommon, and the specific case that would actually break this detector, a local file reachable only via a bare dynamic import and never imported any other way, is rarer still. --> We can do this in another PR though. Its good to keep these PRs lightweight

skhan75 (changes requested):
Looks solid!
Few testing cases needs to be added.

test/unit/core/import-graph.test.ts
BrandonGarate177: I went ahead and added both, multi-line destructure test and a negative test for bare await import() with no assignment. 

Also added the namespaceform (const ns = await import('./m')) since you mentioned it below.

src/core/import-graph.ts
BrandonGarate177: Thanks!

I added the namespace pattern alongside the destructured one, it mirrors the existing require family on lines 60-61. The bare await import('./side-effect.js') form (no assignment) is intentionally not captured, i added a negative test documenting that boundary.

test/unit/core/import-graph.test.ts
```diff
@@ -67,3 +67,55 @@ describe("parseExports — single-name re-export trimming", () => {
...
+
+  it("captures namespace form `const ns = await import('./mod')`", () => {
+    const { names, sources } = parseImports(
+      makeFile(`const utils = await import("./helpers.js");\n`),
+    );
+    expect(names.has("utils")).toBe(true);
+    expect(sources.has("./helpers.js")).toBe(true);
+  });
+
+  it("captures multi-line destructured dynamic imports", () => {
```
skhan75, line 105: 👍

skhan75 (approved):
Great fixes !

PR #32, opened 2026-07-08
Fix Security Consistency detector correctness; rename from "Security Posture"

skhan75:
## What this changes

Hardens the route-level security **consistency** detector and renames the dimension from "Security Posture" to **Security Consistency**, to reflect what it actually measures: whether new code follows the auth / input-validation / rate-limit conventions the rest of the repo already uses, not an absolute security grade.

### Fixes

- **The auth consistency vote now scopes to state-changing routes.** It previously counted read-only `GET` routes in the denominator, which pulled intentionally-public reads into the vote and made the "N of M routes require auth" summary inaccurate.
- **Thin route samples no longer move the score.** A route-consistency finding drawn from too few peer routes to be reliable is surfaced as advisory (informational) instead of affecting the Vibe Drift Score. It still renders, so a small repo is never silently passed.
- **`get_dominant_pattern("auth")` returns the auth convention.** Auth, input-validation, and rate-limit conventions now have separate cached votes, so querying one no longer returns whichever happened to have the widest sample.
- **Consistent advisory rendering.** A below-threshold advisory finding renders the same way across the terminal, HTML, CSV, and DOCX outputs, and never appears as a scored drift finding while the category reads not-applicable.

### Rename

- "Security Posture" is now "Security Consistency" everywhere it is shown (score bar, HTML report card, DOCX export), with a permanent note that consistency is not the same as the absence of vulnerabilities. Internal identifiers and analyzer IDs are unchanged.

### Migration notes

- The scoring formula version is bumped, so the first scan after upgrading suppresses cross-version score deltas (a one-time, silent migration, no per-scan banner).
- The cached baseline format version is bumped; stale baselines rebuild automatically on the next scan.

https://claude.ai/code/session_01XkqH9w2pU7b7Hctbe45h2t

### DIFF
--- a/src/scoring/engine.ts
+++ b/src/scoring/engine.ts
@@ -54,7 +54,7 @@ import {
  * where possible and a one-time release-notes notice is shown (see
  * src/core/scoring-notice.ts). Users never see this string.
  */
-export const SCORING_VERSION = "v7";
+export const SCORING_VERSION = "v8";
 
 /** The bundled corpus distribution, typed. Placeholder until the corpus build lands. */
 export const scorePercentiles = scorePercentilesArtifact as ScorePercentiles;
--- a/src/scoring/engine.ts
+++ b/src/scoring/engine.ts
@@ -129,6 +129,30 @@ export function applySecurityMinPeerFloor(findings: Finding[]): Finding[] {
   return changed ? out : findings;
 }
 
+// The route-consistency drift finding's `driftCategory`: SECURITY_DRIFT_ID
+// (the `Finding.analyzerId`) without the "drift-" prefix. The raw
+// `DriftFindingReport` array keeps this category tag but has no `analyzerId`,
+// so renderers that read `driftFindings` directly test the floor on this
+// instead of on the demoted analyzer id.
+const SECURITY_DRIFT_CATEGORY = "security_posture";
+
+/**
+ * DriftFinding-report view of the same min-peer floor `applySecurityMinPeerFloor`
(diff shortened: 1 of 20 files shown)


src/scoring/engine.ts
```diff
@@ -102,6 +102,57 @@ export function applyReimplementationConcentrationGate(
...
+ * but never dent the Vibe Drift Score. Findings at or above the floor keep the
+ * drift id and score, ramped by groupSampleConfidence up to
+ * SAMPLE_FULL_CONFIDENCE (8). Route-independent taint findings (codedna-taint)
+ * are unaffected. Pure; returns the same array reference when nothing re-tags.
+ */
+export const MIN_SECURITY_PEERS = 4;
+const SECURITY_DRIFT_ID = "drift-security_posture";
+const SECURITY_ADVISORY_ID = "security_posture-advisory";
+
+export function applySecurityMinPeerFloor(findings: Finding[]): Finding[] {
```
skhan75, line 120: FYI - VibeDrift's security check doesn't know what "secure" means in the abstract. It works by majority vote among your own routes: "9 of your 10 mutating routes go through requireAuth, this one doesn't, that's drift." The vote only means something if there are enough voters.

The problem it fixes: the detector can fire with as few as 2 routes. Imagine a tiny service with exactly two POST routes - one has auth, one doesn't. Which one is "the convention" and which one is "the deviation"? You genuinely can't say; it's a coin flip, not a pattern. The old behavior still subtracted score points for it, which means a 200-line hobby repo could get dinged on Security based on a sample of two.

PR #36, opened 2026-07-09
feat(import-graph): replace regex parsing with AST + real module resolution

### LINKED ISSUE #33: dead-code: replace regex import parsing and basename matching with AST-based module resolution
The import graph behind the dead-code and phantom-scaffolding checks has the right architecture. Per-file import and export edges, incoming-edge counts, orphan nodes flagged as dead. The weak part is how the edges get built, on both ends.

**Extraction is regex.** `parseImports` and `parseExports` in `src/core/import-graph.ts` run a battery of regex patterns over file content, even though the project already ships web-tree-sitter for other analyzers. Regexes miss and misparse forms the AST handles for free (multiline imports, comments containing import-like text, unusual spacing, re-export chains).

**Resolution is basename matching.** An `import "./utils.js"` credits an incoming edge to every file named `utils` anywhere in the repo, via `fileBasename` and `sourceLookupKey`. That inflates `incomingCount`, hides genuinely dead files, and is the layer where the Windows separator bug (#27) lived. The `index.ts` parent-directory special case exists only to patch around this.

## Proposal

1. Parse import and export edges from the tree-sitter AST we already bundle, for JS and TS.
2. Resolve import specifiers to actual files: relative paths, extension mapping (`./x.js` importing `x.ts`), `index.*` files, and tsconfig `paths`/`baseUrl` aliases. Bare package specifiers resolve to no file edge.
3. Retire the basename lookup as the primary mechanism.

## Acceptance criteria

- [ ] Import/export edges come from tree-sitter parsing, not `IMPORT_PATTERNS`/`EXPORT_*_PATTERNS` regexes, for JS/TS.
- [ ] Two files with the same basename in different directories never share an edge (regression test for the collision case).
- [ ] Resolution covers: relative specifiers, `.js` -> `.ts`/`.tsx` extension mapping, `index.*` directory imports, tsconfig `paths` aliases, dynamic `import()`.
- [ ] Bare package imports (`react`, `zod`) produce no file edge and no incoming-count credit.
- [ ] Existing Windows backslash regression tests still pass without the `fileBasename` dual-separator workaround being load-bearing.
- [ ] Dead-code and phantom-scaffolding both consume the new resolution; a self-scan of this repo reports no new false positives compared to main.
- [ ] Graph build time stays within 1.5x of the current implementation on a 1k-file repo.

BrandonGarate177:
## Summary

We're replacing the import graph's regex-based extraction and basename matching with tree-sitter AST parsing and real relative-path module resolution. The import graph is used by the `dead-code` and `phantom-scaffolding` detectors to determine which files are connected.

**Problem:** `parseImports`/`parseExports` run regex patterns over the raw file contents, matching inside comments, missing multiline imports, and mishandling unusual spacing. We use basename matching (`fileBasename`/`sourceLookupKey`), which credits an import to every file with the same name anywhere in the repo, hiding genuinely dead files and producing false edges.

**Fix:** Two new modules:
- `src/core/import-graph-ast.ts` 
       - walks the tree-sitter AST to extract import sources, imported names, re-exports, dynamic `import()`, and `require()` calls (including nested inside functions/blocks)
- `src/core/import-resolver.ts`
       - resolves import specifiers to actual file paths using relative path resolution, `.js`→`.ts` extension mapping, `index.*` directory imports, and `@/*` tsconfig path aliases. Handles Windows backslash paths natively.

`buildImportGraph()` uses AST + real resolution as the primary path, falling back to regex + basename matching for files without a parsed tree (non-JS/TS) or specifiers that can't be resolved.

### Results (self-scan)

| Metric | Before (upstream/main) | After (this branch) |
|--------|----------------------|---------------------|
| Unused exports | 65 | **24** |
| Orphan files | 11 | **5** |
| Vibe Drift Score | 91 | 90.4 (more accurate, there were false edges hiding dead code) |

### Performance

Measured with `node bin/vibedrift.mjs --json --local-only --no-cache` on two repos:

**VibeDrift (258 files):**
```
UPSTREAM MAIN:  1807ms | 1751ms | 1797ms  (avg 1785ms)
THIS BRANCH:    1884ms | 1893ms | 1869ms  (avg 1882ms)
Ratio: 1.05×
```

**Prisma (2816 files):**
```
UPSTREAM MAIN:  5813ms | 5763ms | 5773ms  (avg 5783ms)
THIS BRANCH:    6582ms | 6210ms | 7264ms  (avg 6685ms)
Ratio: 1.16×
```

### Files changed

| File | Change |
|------|--------|
| `src/core/import-graph-ast.ts` | New — AST-based `parseImportsAst` + `parseExportsAst` with recursive nested import scanning |
| `src/core/import-resolver.ts` | New — real path resolution with extension mapping, index files, tsconfig aliases, Windows support |
| `src/core/import-graph.ts` | Updated — uses AST+resolver primary, regex+basename fallback; updated module doc; `buildImportGraph` accepts optional `ResolverConfig` |
| `test/unit/core/import-graph.test.ts` | 8 new regression tests |

## Acceptance Criteria (from #33)

| Criterion | Status | Evidence |
|-----------|--------|----------|
| Import/export edges from tree-sitter, not regex, for JS/TS | pass | `buildImportGraph` dispatches to `parseImportsAst`/`parseExportsAst` when `file.tree` exists |
| Two files with same basename never share an edge | pass | Regression test: `src/utils/helpers.ts` gets 1 incoming, `test/helpers.ts` gets 0 |
| Resolution: relative specifiers | pass | `resolveRelativeSource` with `joinPath` + `directoryOf` |
| Resolution: .js → .ts/.tsx extension mapping | pass | `EXTENSION_MAP` + test |
| Resolution: index.* directory imports | pass | `INDEX_FILES` array + test |
| Resolution: tsconfig paths aliases | pass | `resolveAlias` with `@/*` → `src/*` + test |
| Resolution: dynamic import() | pass | `extractNestedImports` recursively finds `import()` calls anywhere in the tree |
| Bare packages produce no file edge | pass | `isBarePackage` returns true → `resolveImportSource` returns null → no edge. Test covers `react`, `@supabase/supabase-js`, `node:fs`, `zod` |
| Windows backslash tests still pass | pass | `fileBasename` splits on `/[/\\]/` + resolver normalizes `importerPath` backslashes |
| Dead-code + phantom-scaffolding consume new resolution | pass | Both call `buildImportGraph(files)` unchanged — results improve (65→24 unused, 11→5 orphans) |
| Graph build time within 1.5× | pass | 1.05× on 258 files, **1.16× on 2816 files** (Prisma repo — meets "1k-file" criterion) |

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation
- [ ] Refactor
- [ ] Test
- [ ] Chore / infra

## Checklist

## Related issues

Closes #33

### DIFF
--- a/src/core/import-graph-ast.ts
+++ b/src/core/import-graph-ast.ts
@@ -1,0 +1,341 @@
+/**
+ * AST-based import/export extraction using tree-sitter.
+ *
+ * Walks the parsed syntax tree to extract import sources, imported names,
+ * and exported symbols — replacing the regex-based parseImports/parseExports
+ * for files that have a parsed tree available.
+ *
+ * Falls back gracefully: callers should check file.tree before calling these
+ * and use the regex versions as fallback when no tree is available.
+ */
+
+import type { Tree, SyntaxNode } from "./types.js";
+import type { FileExport, FileImports } from "./import-graph.js";
+
+/**
+ * Extract imports from a parsed AST tree.
+ * Handles: static imports, dynamic imports (including nested), re-exports, require().
+ */
+export function parseImportsAst(tree: Tree): FileImports {
+  const names = new Set<string>();
+  const sources = new Set<string>();
+  const root = tree.rootNode;
+
+  for (let i = 0; i < root.childCount; i++) {
+    const node = root.child(i)!;
+
+    if (node.type === "import_statement") {
+      extractStaticImport(node, names, sources);
+    } else if (node.type === "export_statement") {
+      // Re-exports: export { X } from "./y" / export * from "./y"
+      extractReExportSource(node, sources);
+    } else if (node.type === "lexical_declaration" || node.type === "variable_declaration") {
+      // Dynamic imports: const { X } = await import("./y")
+      extractDynamicImport(node, names, sources);
+    }
+  }
+
+  // Scan the ENTIRE tree for dynamic imports and require() calls nested inside
+  // functions, if-blocks, etc. that the top-level pass misses.
+  extractNestedImports(root, names, sources);
+
+  return { names, sources };
+}
+
+/**
+ * Extract exports from a parsed AST tree.
+ */
+export function parseExportsAst(tree: Tree, relativePath: string): FileExport[] {
+  const exports: FileExport[] = [];
+  const root = tree.rootNode;
+
+  for (let i = 0; i < root.childCount; i++) {
+    const node = root.child(i)!;
+    if (node.type !== "export_statement") continue;
+
+    const hasDefault = node.children.some((c) => c?.type === "default");
+    const exportClause = findChild(node, "export_clause");
+    const declaration = node.children.find((c) => c?.type.includes("declaration"));
+    const source = findChild(node, "string");
+
+    if (exportClause && !source) {
+      // export { foo, bar } (local re-export without source)
+      extractExportClauseNames(exportClause, relativePath, exports, false);
+    } else if (exportClause && source) {
+      // export { foo, bar } from "./y" — re-export with source
+      // Names are exported from THIS file (the barrel)
+      extractExportClauseNames(exportClause, relativePath, exports, false);
+    } else if (declaration) {
+      // export function hello() {} / export const X = ...
+      const name = extractDeclarationName(declaration);
+      if (name) {
+        const line = node.startPosition.row + 1;
+        exports.push({ name, file: relativePath, line, isDefault: hasDefault });
+      }
+    } else if (hasDefault) {
+      // export default <expression> — try to extract identifier
+      const ident = node.children.find((c) => c?.type === "identifier");
+      if (ident) {
+        const line = node.startPosition.row + 1;
+        exports.push({ name: ident.text, file: relativePath, line, isDefault: true });
+      }
+    } else if (node.children.some((c) => c?.type === "*")) {
+      // export * from "./y" — namespace re-export, no named exports from here
+      // The namespace_export case: export * as stuff from "./y"
+      const nsExport = findChild(node, "namespace_export");
+      if (nsExport) {
+        const ident = findChild(nsExport, "identifier");
+        if (ident) {
+          const line = node.startPosition.row + 1;
+          exports.push({ name: ident.text, file: relativePath, line, isDefault: false });
+        }
+      }
+    }
+  }
+
+  return exports;
+}
+
+// ─── Helpers ─────────────────────────────────────────────────────────
+
+function extractStaticImport(node: SyntaxNode, names: Set<string>, sources: Set<string>): void {
+  // Extract source path
+  const sourceNode = findChild(node, "string");
+  if (!sourceNode) return;
+  const source = extractStringValue(sourceNode);
+  if (source) sources.add(source);
+
+  // Extract imported names from the import clause
+  const importClause = findChild(node, "import_clause");
+  if (!importClause) return;
+
+  for (let i = 0; i < importClause.childCount; i++) {
+    const child = importClause.child(i)!;
+    if (child.type === "named_imports") {
+      // { foo, bar, baz as x }
+      for (let j = 0; j < child.childCount; j++) {
+        const spec = child.child(j)!;
+        if (spec.type === "import_specifier") {
+          // First identifier is the imported name
+          const ident = findChild(spec, "identifier");
+          if (ident) names.add(ident.text);
+        }
+      }
+    } else if (child.type === "namespace_import") {
+      // * as ns
+      const ident = findChild(child, "identifier");
+      if (ident) names.add(ident.text);
+    } else if (child.type === "identifier") {
+      // Default import: import Foo from "./x"
+      names.add(child.text);
+    }
+  }
+}
+
+function extractReExportSource(node: SyntaxNode, sources: Set<string>): void {
+  // Only re-exports have a source string: export { X } from "./y" or export * from "./y"
+  const sourceNode = findChild(node, "string");
+  if (sourceNode) {
+    const source = extractStringValue(sourceNode);
+    if (source) sources.add(source);
+  }
+}
+
+function extractDynamicImport(node: SyntaxNode, names: Set<string>, sources: Set<string>): void {
+  // Walk variable declarators looking for: X = await import("./y")
+  for (let i = 0; i < node.childCount; i++) {
+    const declarator = node.child(i)!;
+    if (declarator.type !== "variable_declarator") continue;
+
+    // Find the await_expression → call_expression with import
+    const awaitExpr = findDescendant(declarator, "await_expression");
+    if (!awaitExpr) continue;
+
+    const callExpr = findChild(awaitExpr, "call_expression");
+    if (!callExpr) continue;
+
+    // Check that it's an import() call
+    const callee = callExpr.child(0);
+    if (!callee || callee.type !== "import") continue;
+
+    // Extract source from arguments
+    const args = findChild(callExpr, "arguments");
+    if (!args) continue;
+    const sourceNode = findChild(args, "string");
+    if (!sourceNode) continue;
+    const source = extractStringValue(sourceNode);
+    if (source) sources.add(source);
+
+    // Extract names from the destructuring pattern (if any)
+    const pattern = declarator.child(0);
+    if (!pattern) continue;
+
+    if (pattern.type === "object_pattern") {
+      // const { x, y } = await import(...)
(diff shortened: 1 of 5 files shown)


skhan75:
@BrandonGarate177  can you rebase your changes on top of latest main

src/core/import-graph-ast.ts
```diff
@@ -0,0 +1,341 @@
...
+  // functions, if-blocks, etc. that the top-level pass misses.
+  extractNestedImports(root, names, sources);
+
+  return { names, sources };
+}
+
+/**
+ * Extract exports from a parsed AST tree.
+ */
+export function parseExportsAst(tree: Tree, relativePath: string): FileExport[] {
```
skhan75, line 45: Can you make sure we have unit tests covering parseExportsAst, as this is the most critical part of the entire solution

test/unit/core/import-graph.test.ts
```diff
@@ -197,3 +197,82 @@ describe("parseImports — dynamic await import()", () => {
     expect(sources.size).toBe(0);
   });
 });
+
+describe("import-resolver — real path resolution", () => {
```
skhan75, line 201: The new tests don't actually test the new AST code. `makeFile` never sets file.tree, so `buildImportGraph` falls back to the old regex path and the "same basename" test passes through the resolver, not parseImportsAst. So the whole import-graph-ast.ts file is untested.

src/core/import-graph-ast.ts
```diff
@@ -0,0 +1,341 @@
...
+      }
+    } else if (hasDefault) {
+      // export default <expression> — try to extract identifier
+      const ident = node.children.find((c) => c?.type === "identifier");
+      if (ident) {
+        const line = node.startPosition.row + 1;
+        exports.push({ name: ident.text, file: relativePath, line, isDefault: true });
+      }
+    } else if (node.children.some((c) => c?.type === "*")) {
+      // export * from "./y" — namespace re-export, no named exports from here
```
skhan75, line 80: What if there is an export like ` export * as util from "./y"`?

skhan75: `export *` → has a direct `*` child, so the gate on line 82 passes, but there's no name to grab, so it records nothing. But  in case of `export * as util`, the `*` is nested inside a `namespace_export` node, so there's no direct `* `child. The gate is false and the branch is skipped, so util is never recorded.

skhan75: make sure we have both positives and negatives test cases to test this redundantly

src/core/import-graph.ts
```diff
@@ -179,17 +181,35 @@ export function fileBasename(filePath: string): string {
...
+    }
   }
 
-  // Index files by their basename lookup-key for source resolution
+  // Build file index for real path resolution
+  const fileIndex = buildFileIndex(files.map((f) => f.relativePath));
+
+  // Resolver config: tsconfig path aliases (caller-supplied or default)
+  const resolverConfig: ResolverConfig = config ?? {
+    pathAliases: { "@/*": "src/*" },
```
skhan75, line 209: The doc says _"tsconfig path-alias support"_ but we don't actually read `tsconfig` anywhere, this line just hardcodes `@/* → src/*`. That's fine for our repo, but it'll silently point at the wrong file in projects where `@/` means something else (Next.js apps often map it to the root). Can we either read paths from `tsconfig`, or just change the wording so we're not claiming something we don't do?

skhan75: Side note - nothing passes the new config argument. Both callers (`dead-code.ts:86`, `phantom-scaffolding.ts:109`) call buildImportGraph(files) with no config, so only the hardcoded default ever runs. Fine as groundwork, just flagging it's currently unused.

src/core/import-graph-ast.ts
```diff
@@ -0,0 +1,341 @@
...
+function extractReExportSource(node: SyntaxNode, sources: Set<string>): void {
+  // Only re-exports have a source string: export { X } from "./y" or export * from "./y"
+  const sourceNode = findChild(node, "string");
+  if (sourceNode) {
+    const source = extractStringValue(sourceNode);
+    if (source) sources.add(source);
+  }
+}
+
+function extractDynamicImport(node: SyntaxNode, names: Set<string>, sources: Set<string>): void {
```
skhan75: I think `extractDynamicImport` can just be deleted. `extractNestedImports` below already catches every import() and require(), including the top-level ones,

BrandonGarate177:
Sweet, i just rebased on latest main. Let me know how this looks

skhan75:
Nice work on this, the AST + real-resolution direction is a solid upgrade. One thing worth handling before merge, on `.tsx` files:

### `.tsx` is parsed with the plain `typescript` grammar, which doesn't understand JSX

In `language.ts`, `.tsx` maps to `"typescript"`, and `ast.ts` loads the `tree-sitter-typescript` grammar (not the separate `tsx` grammar). So when a `.tsx` file contains JSX, the parser hits `<div>...</div>`, can't parse it, and marks that region of the tree as an error (`tree.rootNode.hasError === true`).

This didn't matter before, because the old code found imports with regex, which is a pure text scan and doesn't care whether the file parsed cleanly. This PR reads the parse tree instead, so anything sitting inside a broken/error region can get dropped.

### It's a real miss, not just theoretical

I parsed a realistic component with the current grammar:

```tsx
import { useState } from "react";              // (1) top of file

export default function Panel() {
  const load = () => { import("./HeavyView").then(...); };   // (2) in a function body
  return (
    <div onClick={load}>
      {View ? <View /> : <Spinner src={import("./Spinner")} />}   // (3) inside JSX
    </div>
  );
}
```

Result:
- ✅ `import ... from "react"` — found
- ✅ `import("./HeavyView")` — found (it sits above the JSX)
- ❌ `import("./Spinner")` — **missed**, because it lives inside the JSX error region

The old regex path caught both dynamic imports. So on `.tsx` files this can silently drop an import edge, which could flag a file as dead/orphan when it actually isn't.

### Scope and suggested fix

Scope is narrow: only `.tsx` files, and only dynamic `import()`/`require()` calls that sit *inside* JSX. Static top-of-file imports are unaffected. So not a hard blocker, but React projects are a big chunk of users, so worth closing.

Fix: load the `tsx` grammar for `.tsx` files. `tree-sitter-wasms` already ships `tree-sitter-tsx.wasm` alongside the typescript one, so it's mostly a matter of routing `.tsx` (and `.jsx` is already fine on the JS grammar) to the right grammar so JSX parses cleanly and the tree has no error regions. A `.tsx` regression test with a dynamic import inside JSX would lock it in.

---
<sub>🤖 Posted with Claude Code on behalf of @skhan75</sub>

skhan75 (changes requested):

src/core/import-graph-ast.ts
BrandonGarate177: Added 5 dedicated tests for `parseExportsAst`: named function/const/class exports, default exports, export clause with aliases, `export * as namespace`, and a negative test for bare `export *` producing no names. All use real `parseFile()` trees.

test/unit/core/import-graph.test.ts
BrandonGarate177: I added 11 tests using real `parseFile()` trees that exercise `parseImportsAst`, `parseExportsAst`, and `buildImportGraph` with `file.tree` populated. The AST path is now directly tested.

src/core/import-graph-ast.ts
BrandonGarate177: The gate now checks `node.children.some((c) => c?.type === "*") || findChild(node, "namespace_export")` so both `export *` and `export * as X` are handled. Added positive and negative tests for both forms.

src/core/import-graph-ast.ts
BrandonGarate177: Ok I deleted it. `extractNestedImports` covers all dynamic `import()` and `require()` calls recursively, including top-level ones.

src/core/import-graph.ts
BrandonGarate177: I updated the wording to "configurable path aliases (defaults to @/* → src/*)

Both callers get the default which works for this repo. When tsconfig reading is added, they'll pass it through.

BrandonGarate177:
`.tsx` files are now loading `tree-sitter-tsx.wasm` through a filePath check in `getLanguage()`. Dynamic imports inside JSX parse cleanly now. Also caught and fixed `export class` not being extracted during testing

skhan75 (approved):
solid work!

Issue #34, opened 2026-07-08
security consistency: post-merge review follow-ups (4 release blockers)


skhan75:
Post-merge review of #32 surfaced a set of defects. The four in the first group are detection regressions or wrong-answer paths and should block the next release. The rest are consistency and copy fixes in the same areas #32 touched.

## Release blockers

- [ ] **Auth vote drops routes with method `ANY`/`ALL`.** The rescope to `MUTATION_METHODS` excludes routes whose method the extractor could not pin down (`security-consistency.ts:213` defaults to `ANY`; `.all()` maps to `ALL`). Flask `@app.route(...)` and Express `.all()` repos lose auth-consistency detection entirely, with no advisory. Treat unknown-method routes as part of the auth peer group, and extend the method extraction.
- [ ] **MCP serves pre-v2 baselines forever.** `getBaseline` never version-checks an existing baseline, so `get_dominant_pattern("auth")` on an old cache finds no `securitySubVotes` and answers the no-vote projection ("consistent, no deviations detected") even when the stale vote recorded auth deviators. `BASELINE_VERSION` in the key also makes the status read "stale" on every call until a full scan runs. Version-gate the load and rebuild on mismatch.
- [ ] **`check_file_drift` still reads the collided `perCategoryVote` slot.** With auth's denominator shrunk by the rescope, the rate-limiting vote (counted over all routes) wins the widest-denominator tie-break more often, so the tool can report `fits: true` for the repo's one unauthed mutating route. Route sub-checks should consult `securitySubVotes` deviators.
- [ ] **`securitySubVotes` bypass the min-peer floor.** They are built from raw findings, so the MCP serves a below-floor auth vote (for example 2 of 3 routes) as the authoritative convention while the same scan demotes that sample as too thin to score. Apply the floor, or attach an explicit thin-sample caveat to the projection.

## Consistency follow-ups

- [ ] `context.md` "Recent trajectory" is not gated on `scoringVersion` (`diffScans` only checks schema version), so the first v8 scan commits cross-version score deltas and re-tag churn into a committed file. The todo entry in #32 assumed this case was already silenced; it is not for this surface.
- [ ] When every security finding is below floor, `scoredDriftView` credits the empty category full health (14/14, 0 findings) and that value is uploaded in `result_json.driftScores` next to the composite's N/A.
[...]

PR #38, opened 2026-07-09
fix(security): close the four #34 release blockers (auth peer group, baseline version gate, sub-vote floor, check_file_drift)

skhan75:
Fixes the four release blockers from #34, on top of merged Phase 2 (#37).

1. **Method-`ANY`/`ALL` routes stay in the auth peer group.** The auth consistency vote and the uniform-auth-gap fallback now share `AUTH_VOTE_METHODS` (mutating + unresolved-method + `.all()` routes), restoring auth detection for Flask-style decorators and Express `.all()`. Validation and rate-limit votes are unchanged. Messages counting unresolved-method routes no longer claim they all mutate.

2. **Persisted baselines carry a version, and the MCP rebuilds stale ones.** `getBaseline` treats a baseline from an older `BASELINE_VERSION` exactly like the never-scanned case: one lazy rebuild, persisted, served fresh — instead of serving pre-v2 votes forever with a permanent "stale" status.

3. **`check_file_drift` consults the security sub-votes.** A file deviating in any of Auth middleware / Input validation / Rate limiting is `fits:false` citing that sub-convention, route-denominated — instead of reading the collided widest-denominator slot that rate limiting usually wins.

4. **`securitySubVotes` respect the min-peer floor.** Sub-votes below `MIN_SECURITY_PEERS` are no longer persisted, so the MCP never serves a thin sample as the repo's authoritative convention while the same scan demotes it as too small to score. The in-loop auth check goes honest-silent on such repos.

`BASELINE_VERSION` bumps 2 → 3 so already-scanned repos rebuild their votes lazily (blockers 1 and 4 change what the votes contain).

16 new tests, red-first; 894 passing total; tsc and lint clean.

Remaining #34 checklist items (consistency + copy fixes) are follow-ups, not blockers.

### DIFF
--- a/src/core/baseline.ts
+++ b/src/core/baseline.ts
@@ -53,6 +53,12 @@ export interface MinhashEntry {
 
 export interface RepoDriftBaseline {
   key: string;
+  /** BASELINE_VERSION this baseline was built with. Optional because baselines
+   *  persisted before the field shipped carry none; loaders treat absence as
+   *  pre-current and rebuild rather than serve (getBaseline's version gate).
+   *  assembleBaseline always sets it, so every newly persisted baseline is
+   *  versioned. */
+  version?: number;
   rootDir: string;
   ctxFiles: Array<{ path: string; hash: string }>;
   perCategoryVote: Partial<Record<DriftCategory, CategoryVote>>;
@@ -187,6 +193,7 @@ export function assembleBaseline(
 
   return {
     key: computeBaselineKey(ctxFiles),
+    version: BASELINE_VERSION,
     rootDir,
     ctxFiles,
     perCategoryVote: votesFromFindings(driftFindings),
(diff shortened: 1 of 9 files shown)


skhan75:
Superseded by #39 (branch \`security-blockers-34\`), which closes the same four blockers with a more precise B1: it resolves each route's HTTP method (parsing Flask \`methods=\`, defaulting a bare \`@app.route\` to GET, treating \`.all()\` as mutating) instead of adding unresolved \`ANY\` routes to the vote, so read-only Flask routes are never flagged as missing auth. #39 also carries the scoring-version migration for the auth-vote change, so the composite shift ships as a versioned refinement with a one-time notice rather than a silent delta. Closing in favor of #39; the intent here is preserved with the added precision.

PR #40, opened 2026-07-09
test: fix 13 Windows-incompatible test assumptions

BrandonGarate177:
## Summary

13 tests fail on Windows due to platform-specific assumptions in the test harness. This fixes all 13 without changing any product code.

**Root causes:**
1. `GIT_AUTHOR_DATE="..." git add` is Unix inline env var syntax — doesn't work on Windows cmd/PowerShell (7 tests)
2. `process.env.HOME = tmpDir` doesn't redirect `os.homedir()` on Windows which uses `USERPROFILE` — tests read stale cache from real home dir (5 tests)
3. `statSync(path).mode & 0o100` checks Unix execute permission bits which don't exist on Windows (1 test)

**Fixes:**
- Use Node's `execP` with `{ env }` option instead of shell inline vars — works on all platforms
- Set both `HOME` and `USERPROFILE` when redirecting home — Linux/macOS reads `HOME`, Windows reads `USERPROFILE`, setting both is harmless
- Skip the permission bit assertion on Windows (`process.platform !== "win32"`) — content assertions still run
- Replace `/dev/null` hooksPath with `.git/hooks-disabled` (nonexistent dir on all platforms)
- Use separate `fakeHome` dir for git-metadata cache isolation (prevents cache files from being tracked by git inside the test repo)
- Add `vi.resetModules()` + dynamic import for `collectGitMetadata` since `CACHE_DIR` is evaluated at module load time

**Before:** 13 failures on Windows, 0 on CI (Ubuntu)
**After:** 0 failures on all platforms

### Score verification

Verified on Windows 11 (Node 24) that the CLI scan score is unchanged:

```
Base commit (d89137f — the main this PR branches from): 90.8
This branch (fix/windows-test-compat):                  90.8
```

Score is identical — only test files are modified, zero product code touched.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Refactor
- [x] Test
- [ ] Chore / infra

## Checklist

## Related issues

No tracked issue — discovered while contributing on Windows.

### DIFF
--- a/test/integration/mcp.test.ts
+++ b/test/integration/mcp.test.ts
@@ -33,7 +33,7 @@ async function connect(home: string): Promise<Client> {
   const transport = new StdioClientTransport({
     command: "node",
     args: [CLI, "mcp"],
-    env: { ...process.env, HOME: home },
+    env: { ...process.env, HOME: home, USERPROFILE: home },
   });
   const client = new Client({ name: "test-client", version: "0.0.0" });
   await client.connect(transport);
--- a/test/integration/mcp.test.ts
+++ b/test/integration/mcp.test.ts
@@ -56,7 +56,7 @@ describe("MCP integration — Pro plan (tools serve)", () => {
     // Build the baseline the real way: a scan under the SAME HOME the server reads
     // (the cache dir is derived from HOME, so this writes to home/.vibedrift).
     execFileSync("node", [CLI, repo, "--local-only", "--format", "terminal"], {
-      env: { ...process.env, HOME: home },
+      env: { ...process.env, HOME: home, USERPROFILE: home },
       stdio: "ignore",
     });
     client = await connect(home);
@@ -155,7 +155,7 @@ describe("MCP integration — deep-scan nudge piggybacks on a write-time tool", 
     }
     writeFileSync(join(repo, "then0.ts"), "export function then0(){\n  return a().then(r => r);\n}\n");
     execFileSync("node", [CLI, repo, "--local-only", "--format", "terminal"], {
-      env: { ...process.env, HOME: home },
+      env: { ...process.env, HOME: home, USERPROFILE: home },
       stdio: "ignore",
     });
     client = await connect(home);
--- a/test/integration/temporal-pivot.test.ts
+++ b/test/integration/temporal-pivot.test.ts
@@ -17,13 +17,13 @@ async function gitInit(dir: string): Promise<void> {
   await run(dir, "git init -q");
   await run(dir, 'git config user.email "test@example.com"');
   await run(dir, 'git config user.name "Test"');
-  await run(dir, "git config core.hooksPath /dev/null");
+  await run(dir, "git config core.hooksPath .git/hooks-disabled");
 }
 
 async function commit(dir: string, isoDate: string, msg: string): Promise<void> {
-  const env = `GIT_AUTHOR_DATE="${isoDate}" GIT_COMMITTER_DATE="${isoDate}"`;
-  await run(dir, `${env} git add -A`);
-  await run(dir, `${env} git commit -q -m "${msg}"`);
+  const env = { ...process.env, GIT_AUTHOR_DATE: isoDate, GIT_COMMITTER_DATE: isoDate };
+  await execP("git add -A", { cwd: dir, env });
+  await execP(`git commit -q -m "${msg}"`, { cwd: dir, env });
 }
 
 /**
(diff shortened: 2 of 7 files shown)


anishekkamal (changes requested):
Thanks for this — verified end-to-end on a real Windows machine (Windows 11, Node 24):

- `main`: exactly the 13 failures you describe, in exactly these 5 files. ✅
- This branch: **900/900 pass**, plus `lint`, `typecheck`, and `build` all green. ✅
- All three root-cause diagnoses check out (inline env-var syntax vs `cmd.exe`, `os.homedir()` reading `USERPROFILE` on Windows, no POSIX exec bits). The `vi.resetModules()` + dynamic-import pattern for the module-load-time `CACHE_DIR` is the correct, non-obvious fix — nice diagnosis.

One in-scope gap found while testing, which is why this is "request changes" rather than approve:

### `test/integration/mcp.test.ts` — the two `execFileSync` baseline scans still set only `HOME`

The fix was applied to `connect()` but not to the two child-process scans at lines ~58 and ~157:

```ts
execFileSync("node", [CLI, repo, "--local-only", "--format", "terminal"], {
  env: { ...process.env, HOME: home },   // ← needs USERPROFILE: home too
  stdio: "ignore",
});
```

On Windows the spawned CLI inherits the real `USERPROFILE` via `...process.env`, so the baseline scan writes to the developer's **real** `~/.vibedrift` (confirmed empirically — my real `~/.vibedrift/baseline-cache` gained files timestamped exactly at the test run). Meanwhile the MCP server (whose env *is* fixed) reads the fake home, finds no baseline, and lazily re-indexes — you can see `vibedrift-mcp: indexing … for the first time` in the test output. The tests still pass because the assertions accept `ok|stale`, but on Windows they exercise the re-index path instead of the intended pre-built-baseline path, and they pollute the real home dir. Same one-line fix in both spots:

```ts
env: { ...process.env, HOME: home, USERPROFILE: home },
```

### Optional (pre-existing, outside your diff — happy to take as follow-up instead)

Two sibling tests have the identical `HOME`-only pattern and silently write to the real `~/.vibedrift` on Windows (they pass either way, so your "0 failures" claim still holds):

- `test/unit/core/embedding-index.test.ts` (~line 79) — `beforeAll` sets only `process.env.HOME`; `indexDir()` calls `homedir()` lazily, so setting/restoring `USERPROFILE` alongside `HOME` is sufficient (no `resetModules` needed).
- `test/unit/ml-client/build-embedding-index.test.ts` (~line 23) — same pattern, same fix.

If you'd rather keep this PR tight, fix just `mcp.test.ts` and we'll file the other two as a follow-up issue.

BrandonGarate177:
I added USERPROFILE: home to both execFileSync baseline scans in mcp.test.ts (lines 58 and 157) and to the build-embedding-index.test.ts setup. Same pattern as the connect() fix already in this PR. 

Thanks for catching the pollution to ~/.vibedrift , cause I definitely wouldn't have caught that. Both callers now write to the fake home so the MCP server exercises the intended pre-built-baseline path instead of re-indexing from scratch.

skhan75 (dismissed):
Nice sweep on the HOME/USERPROFILE fixes

anishekkamal (changes requested):
Thanks @BrandonGarate177 - verified on Windows 11 / Node 24, full suite 900/900 green. The `mcp.test.ts` USERPROFILE additions and the `build-embedding-index.test.ts` fix (nice touch deleting the vars on restore when they were previously undefined) both look correct.

One more instance of the same pattern slipped through: `test/unit/core/embedding-index.test.ts` (the "write/load round-trip" block) still sets only `HOME`. Since `src/core/embedding-index.ts` resolves the cache via `homedir()` — which ignores `HOME` and reads `USERPROFILE` on Windows — this test writes to the developer's real `~/.vibedrift/embedding-index` instead of the temp dir. I confirmed it: cleaned my home dir, ran this test alone, and a stray `*.json` landed in the real `~/.vibedrift/embedding-index`. The block's own comment ("the test never touches the real ~/.vibedrift") is inaccurate on Windows.

Same fix as the others:

```ts
  let prevHome: string | undefined;
  let prevUserProfile: string | undefined;
  beforeAll(() => {
    home = mkdtempSync(join(tmpdir(), "vd-eidx-"));
    prevHome = process.env.HOME;
    prevUserProfile = process.env.USERPROFILE;
    process.env.HOME = home;
    process.env.USERPROFILE = home;
  });
  afterAll(() => {
    if (prevHome !== undefined) process.env.HOME = prevHome;
    else delete process.env.HOME;
    if (prevUserProfile !== undefined) process.env.USERPROFILE = prevUserProfile;
    else delete process.env.USERPROFILE;
    rmSync(home, { recursive: true, force: true });
  });


With that last one in, this is good to merge. Everything else is solid.

BrandonGarate177:
We added `USERPROFILE` to the `embedding-index.test.ts` `beforeAL / afterAll` block with the same save / restore pattern. The test no longer writes to the real `~/.vibedrift/embedding-index` on windows

skhan75 (approved):

PR #65, opened 2026-07-17
fix(output): three honesty fixes in scan output

### LINKED ISSUE #64: Four small honesty fixes in scan output (contributor-friendly batch)
VibeDrift promises honest output. These are four small, self-contained places where the output says something that isn't quite true. Each one can be picked up on its own, and none of them needs deep knowledge of the codebase. If you want to grab just one, comment which one and go for it.

- [x] 1. Watch mode tells signed-in users to sign in
- [x] 2. The "since last scan" diff can call something drift that the report itself calls advisory
- [x] 3. The calibration README describes a harness that no longer exists in that shape
- [ ] 4. The suppression fallback can silently hide a route on a malformed line

---

## 1. Watch mode tells signed-in users to sign in

**What happens.** `vibedrift watch` requires an account and exits if you have no token. Yet every rescan it prints ends with a nudge that says

```
· Sign in with `vibedrift login` (free, 10s) and run `vibedrift --deep`. 1 free deep scan/month.
```

The user it is talking to is already signed in. The same wrong nudge appears on any signed-in `vibedrift . --format terminal` scan.

Image

*The screenrecording shows `vibedrift status` reporting an authenticated session, then `vibedrift watch` ending its output with the sign-in ask.*

**Why it happens.** The nudge copy is built by `generateTeaseMessages` in `src/output/tease.ts` (the sign-in line is appended near line 155), and that function never learns whether the session is authenticated. The scan command calls it in `src/cli/commands/scan.ts` (around line 471) without passing any auth state. Watch mode also runs its scans with `localOnly: true`, so nothing downstream can recover the auth state on its own.

**Suggested direction.** Thread a simple boolean (bearer token present) into `generateTeaseMessages` and pick the closing line based on it. A signed-in user should see something like "Run `vibedrift . --deep` to reveal them. Your first deep scan each month is free." with no sign-in ask. The banner rendered in `src/output/terminal.ts` (the Deep Analysis Preview block) uses whatever messages it receives, so the renderer itself likely needs no change.

**Acceptance criteria**

- [ ] A signed-in `vibedrift watch` session never prints "Sign in with `vibedrift login`"
- [ ] A signed-in `vibedrift . --format terminal` scan never prints it either
- [ ] A signed-out scan keeps the current sign-in copy unchanged
- [ ] A regression test covers both branches of the copy

---

## 2. The "since last scan" diff can call something drift that the report itself calls advisory

**What happens.** VibeDrift demotes thin security findings to "advisory" before scoring, so a finding based on too few routes never counts as drift. The report renders it in the hygiene section and the category shows N/A. But the scan-over-scan comparison (the "Since last scan" banner and the "Recent trajectory" section of the committed `.vibedrift/context.md`) is computed from the raw findings, before that demotion. So a newly introduced below-floor finding can show up in the diff as a "new drift finding" while the very same report renders it as advisory. The suppression audit note (the "N routes excluded" bookkeeping finding) leaks into the diff the same way.

**Example.** Your repo has two mutating routes and one lacks auth. That is below the peer floor, so the report says Security Consistency N/A and shows the finding as advisory. Run a second scan after introducing it and the diff banner may still announce a new drift finding, contradicting the page it is printed on.

Image

**Where to look.** In `src/cli/commands/scan.ts` the diff digests are built from `driftResult.driftFindings` (the raw set), while the rendered view comes from `scoredDriftView(...)` a few lines later. The fix is to feed the diff digest and `saveScanResult` from the scored view, and to do both together. If the two sources differ, every scan reports a one-time spurious diff against its own history.

**One thing not to touch.** The baseline (`assembleBaseline`) deliberately reads the raw representation. Leave it that way.

**Acceptance criteria**

- [ ] A below-floor advisory finding introduced between two same-version scans does not appear in the diff banner as new drift
- [ ] Adding or removing a `@vibedrift-public` suppression annotation does not appear in the diff as a new or resolved finding
- [ ] `.vibedrift/context.md`'s trajectory section stays consistent with the rendered report
- [ ] The saved scan and the diff digest are built from the same source, covered by a test
- [ ] Re-scanning an unchanged repo after the fix reports no changes

---

## 3. The calibration README describes a harness that no longer exists in that shape

**What happens.** `test/calibration/README.md` makes two claims the code has moved past.

1. The sample output says the monotonic gate requires a score drop of at least 5 points per 25% injected drift. The actual gate in `test/calibration/run.ts` is `REQUIRED_DROP_PER_25PCT = 3.0`.
2. The "add your own scenario" section says to drop a generator file in a `generators/` directory. No such directory exists. Injectors live in `test/calibration/injectors.ts` as entries in a single file.

Image

**Why it matters.** This README is the front door for anyone trying to extend the calibration harness. Following it as written leads to a directory that does not exist and expectations the gate will not enforce.

**Acceptance criteria**

- [ ] The README's numbers match the constants in `run.ts` (quote the constant rather than hardcoding the value twice, or state it as "see `REQUIRED_DROP_PER_25PCT`")
- [ ] The "add a scenario" instructions describe `injectors.ts` as it works today
- [ ] No other stale reference remains in that README (worth a quick read-through while in there)

---

All four have their current behavior verified against the source before this issue was filed. If you pick one up and something above does not match what you see, say so here rather than working around it.

BrandonGarate177:
## Summary

Three honesty fixes in scan output, addressing items 1–3 of #64.

**1. Watch mode tells signed-in users to sign in**
The sign-in nudge is now gated on whether the session is actually authenticated. The catch the first pass missed: `bearerToken` is hard-set to `null` under `--local-only`, and `vibedrift watch` always scans local-only — so a signed-in watch session still printed the nudge. Fixed by resolving signed-in state independently of the network in `runScan`:

```ts
const signedIn = bearerToken !== null || (await resolveToken()) !== null;
```

`resolveToken()` is purely local (flag → env → config file, zero egress), so this doesn't break the `--local-only` promise. `signedIn` is threaded through `buildScanResult` into `generateTeaseMessages`; every network path still gates on the real `bearerToken`. Signed-in users now see "Run `vibedrift . --deep` to reveal them. Your first deep scan each month is free." with no sign-in ask; signed-out copy is unchanged.

**2. "Since last scan" diff calls advisory findings "drift"**
The diff digest was built from `driftResult.driftFindings` (raw, before advisory demotion), so a below-floor security finding could appear as "new drift" in the diff banner while the report rendered it as advisory/N/A. `scoredDriftView(...)` now runs above the diff, and both the diff digest and `saveScanResult` are fed from the scored view. The baseline (`assembleBaseline`) still reads the raw representation unchanged.

**3. Calibration README describes a harness that no longer exists**
- Sample output claimed `≥5pt` drop per 25% injection; actual gate is `REQUIRED_DROP_PER_25PCT = 3.0`
- "Adding a new injection type" referenced a `generators/` directory that doesn't exist; injectors live in `test/calibration/injectors.ts` with a `BaselineFile[]` signature

## Changes since review

Addressing @skhan75's review:

- **Watch path now actually fixed (blocking 1).** Switched from `bearerToken !== null` to the local `resolveToken()`-based `signedIn` above, so a signed-in `vibedrift watch` session no longer prints the nudge — the first acceptance criterion. Confirmed the path: `watch` → `runScan(localOnly: true)` → `buildScanResult` → `generateTeaseMessages(signedIn)`.
- **Regression tests added (blocking 2).**
  - `test/unit/output/tease.test.ts` — both copy branches (signed-in never contains "Sign in with"; signed-out unchanged; defaults to signed-out; deep scan emits no tease).
  - `test/unit/output/scan-diff-advisory.test.ts` — `scoredDriftView` drops below-floor security + suppression findings; a below-floor advisory introduced between scans is not "new" drift; a `@vibedrift-public` suppression is neither "new" nor "resolved"; the saved scan and diff digest come from the same scored source.
- **Stale comment removed (blocking 3).** The old RAW-`driftFindings` comment above `saveScanResult` contradicted the new code; removed the outdated half so it tells one story.
- **CHANGELOG.** Added Unreleased notes for the sign-in and diff/advisory fixes, including the one-time "resolved" self-heal on the first scan after upgrade.

One deliberate choice on your non-blocking note: I kept the signed-in closing line (it carries the "first deep scan free" CTA) rather than dropping it in favor of the banner — happy to switch to the no-line approach if you'd prefer.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Refactor
- [ ] Test
- [ ] Chore / infra

## Checklist

## Related issues

Closes #64 (items 1, 2, 3 — item 4 shipped in #66)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,17 @@ explicitly under **Breaking** so CI users can recalibrate.
   the dashboard) omits it instead of showing a full-score bar next to the N/A.
 - Removed a crash risk on extremely large finding sets (whole-array argument spreads
   replaced with loops).
+- **Signed-in users no longer see the sign-in nudge.** `vibedrift watch` and any signed-in
+  `--format terminal` scan previously ended with "Sign in with `vibedrift login`" even
+  though the session was already authenticated. Signed-in state is now resolved locally
+  (zero egress — `--local-only` and `watch` are unaffected) and the closing line reads
+  "Run `vibedrift . --deep` to reveal them. Your first deep scan each month is free."
+- **The "Since last scan" diff no longer calls advisory findings drift.** The diff digest
+  and the persisted scan are both built from the scored drift view, so a below-floor
+  security finding (or a `@vibedrift-public` suppression) that the report renders as
+  advisory can no longer appear as a "new drift finding" in the banner or in
+  `.vibedrift/context.md`. Note: the first scan after this upgrade on a repo whose history
+  already persisted such findings reports them as "resolved" once, then self-heals.
 
 ## 0.16.2 — 2026-07-16
 
--- a/src/cli/commands/scan.ts
+++ b/src/cli/commands/scan.ts
@@ -389,6 +389,7 @@ async function buildScanResult(
   startTime: number,
   timings: Record<string, number>,
   bearerToken: string | null,
+  signedIn: boolean,
   apiUrl: string | undefined,
   spinner: ReturnType<typeof ora> | null,
 ): Promise<ScanResult> {
@@ -468,7 +469,7 @@ async function buildScanResult(
   // Tease — show "run --deep" upsell only when user is *not* using deep mode.
   // Pass codeDnaResult so the tease can name specific near-duplicate files
   // and opaque-named functions deep scan would confirm, not generic copy.
-  const teaseMessages = generateTeaseMessages(ctx, allFindings, options.deep === true, codeDnaResult, bearerToken !== null);
+  const teaseMessages = generateTeaseMessages(ctx, allFindings, options.deep === true, codeDnaResult, signedIn);
   // Free Tier-1 reimplementation teaser (count only). Skipped on deep scans —
   // there the real panel-confirmed ml-reimplementation findings render instead.
   const reimplementationCandidates = options.deep
@@ -1135,6 +1136,15 @@ export async function runScan(
     ? await resolveAuthAndBanner(options)
     : { bearerToken: null, apiUrl: undefined };
 
+  // Signed-in state is independent of network mode. bearerToken is hard-set to
+  // null under --local-only (and `vibedrift watch` always scans local-only), so
+  // it cannot answer "is this user signed in?". resolveToken() is purely local
+  // (flag → env → config file, zero egress), so we can detect a signed-in
+  // session without breaking the --local-only zero-egress promise. This drives
+  // the tease copy only; every network path stays gated on the real bearerToken.
+  // (#64 item 1: a signed-in watch session must not print the sign-in nudge.)
+  const signedIn = bearerToken !== null || (await resolveToken()) !== null;
+
   // Kick off the passive update check as soon as we know the network
   // is allowed. Runs in parallel with the scan so it adds zero latency
   // in the common case (cache hit) and ~200ms worst-case (uncached
@@ -1170,7 +1180,7 @@ export async function runScan(
 
   const timings = { ...pipeline.timings, ...deepTimings.timings };
 
-  const result = await buildScanResult(pipeline, options, startTime, timings, bearerToken, apiUrl, spinner);
+  const result = await buildScanResult(pipeline, options, startTime, timings, bearerToken, signedIn, apiUrl, spinner);
 
   // Persist the RepoDriftBaseline as a scan side-effect so the MCP server's
   // cold start is a fast load rather than a 3–8s rebuild. Reuses the ctx +
@@ -1294,14 +1304,11 @@ export async function runScan(
   // Persisting `scoringVersion` lets future scans detect that this scan was
   // computed under a different formula and skip cross-version deltas.
   //
-  // Use the RAW driftResult.driftFindings (not the rendered/filtered
-  // result.driftFindings) so the persisted drift digests match the ones the
-  // scan-over-scan diff computes from the same raw array (see buildScanResult).
-  // If these two sources diverged, a below-floor security finding present in
-  // one but not the other would show as a spurious "new/resolved drift finding"
   // Save using the scored (rendered) drift view so the persisted scan matches
-  // the diff digest. A below-floor advisory never produces a spurious diff.
-  // The baseline (assembleBaseline) still reads the raw representation.
+  // the diff digest the scan-over-scan comparison computes from the same view.
+  // A below-floor advisory finding therefore never shows as a spurious
+  // "new/resolved drift finding". The baseline (assembleBaseline) still reads
+  // the raw representation.
   await saveScanResult(
     rootDir,
     result.scores,
--- a/src/cli/commands/scan.ts
+++ b/src/cli/commands/scan.ts
@@ -468,7 +468,7 @@ async function buildScanResult(
   // Tease — show "run --deep" upsell only when user is *not* using deep mode.
   // Pass codeDnaResult so the tease can name specific near-duplicate files
   // and opaque-named functions deep scan would confirm, not generic copy.
-  const teaseMessages = generateTeaseMessages(ctx, allFindings, options.deep === true, codeDnaResult);
+  const teaseMessages = generateTeaseMessages(ctx, allFindings, options.deep === true, codeDnaResult, bearerToken !== null);
   // Free Tier-1 reimplementation teaser (count only). Skipped on deep scans —
   // there the real panel-confirmed ml-reimplementation findings render instead.
   const reimplementationCandidates = options.deep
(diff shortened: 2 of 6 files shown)


skhan75 (changes requested):
Thanks for picking this up so quickly, and for following the issue's prescription closely. Item 2's mechanics are exactly right (digest and save move together, `assembleBaseline` stays on the raw representation), and both README fixes check out against the code. Three things need another pass before this can merge.

## 1. The watch path is not actually fixed yet (the headline of item 1)

`bearerToken !== null` looks right at the call site, but `bearerToken` is only resolved when the network is enabled. Under `--local-only` it is hard-set to `null` (`src/cli/commands/scan.ts`, around line 1136)

```ts
const networkEnabled = !options.localOnly;
const { bearerToken, apiUrl } = networkEnabled
  ? await resolveAuthAndBanner(options)
  : { bearerToken: null, apiUrl: undefined };
```

and `vibedrift watch` always runs its scans with `localOnly: true`. So in every watch rescan `isSignedIn` is `false` and the sign-in copy still prints. Your change fixes the signed-in `vibedrift . --format terminal` case, which is real progress, but the first acceptance criterion on the issue (a signed-in watch session never prints the sign-in ask) still fails.

The good news is the fix is cheap. `resolveToken()` is purely local (flag, then env, then config file, zero network, see `src/auth/resolver.ts` around line 24), and `scan.ts` already calls it under `--local-only` for the `--write-context` gate around line 1103. So you can resolve the signed-in state independently of `networkEnabled` without breaking the zero-egress promise of `--local-only`. Something like

```ts
const signedIn = bearerToken !== null || (await resolveToken()) !== null;
```

and pass `signedIn` through.

## 2. Regression tests are missing

The checklist says new behavior has tests, but the diff contains no test files. The issue's acceptance criteria ask for them on items 1 and 2, and repo convention is that bug fixes include a regression test. The minimum that would pin this

- both branches of the tease copy (signed-in output never contains "Sign in with", signed-out copy unchanged)
- a below-floor advisory finding introduced between two scans does not appear as new drift in the diff
- the saved scan and the diff digest are built from the same source

`test/unit/output/` has fixtures you can crib from (`security-advisory-render.test.ts` builds below-floor security findings already).

## 3. The comment above `saveScanResult` now says the opposite of the code

On your branch the old comment's first half survives (around line 1297)

```
// Use the RAW driftResult.driftFindings (not the rendered/filtered
// result.driftFindings) so the persisted drift digests match ...
// If these two sources diverged, a below-floor security finding present in
// one but not the other would show as a spurious "new/resolved drift finding"
```

immediately followed by your new lines saying the reverse. A future reader following the stale half would revert your fix. Please remove the outdated part so the comment tells one story.

## Two smaller thoughts, not blocking

- The new signed-in closing line nearly duplicates the banner the terminal renderer prints right after it ("Run vibedrift . --deep to reveal them. 1 free deep scan/month."). An option worth considering is pushing no closing line at all when signed in and letting the banner carry the CTA. As a bonus, the "N additional AI findings" count would stop counting the CTA line as a finding.
- Because previously saved scans persist raw digests, the first scan after this change on a repo that has below-floor advisory findings will report those findings as "resolved" once, then self-heal. Worth a one-line note in the CHANGELOG Unreleased section.

Really appreciate the contribution. With the watch path covered and tests in place this is a clean close of three issue items.

🤖 Review drafted with [Claude Code](https://claude.com/claude-code)

BrandonGarate177:
Pushed a new commit addressing all three:

**1. Watch path.** You were right, `bearerToken` is null under `--local-only` and watch always scans local-only, so the old `bearerToken !== null` never saw the signed-in state. Switched to your suggestion:

```ts
const signedIn = bearerToken !== null || (await resolveToken()) !== null;
```

and threaded `signedIn` through `buildScanResult` into the tease. Traced it end to end — `watch` → `runScan(localOnly: true)` → `buildScanResult` → `generateTeaseMessages(signedIn)` — so a signed-in watch session no longer prints the nudge, and the deep-AI network paths still gate on the real `bearerToken` so `--local-only` stays zero-egress.

**2. Tests.** Added the ones you listed plus a bit more:
- `tease.test.ts` — both copy branches (signed-in never says "Sign in with", signed-out unchanged, defaults to signed-out, deep = no tease).
- `scan-diff-advisory.test.ts` — a below-floor advisory introduced between two scans isn't "new" drift, a `@vibedrift-public` suppression is neither new nor resolved, and the saved scan + diff digest come from the same scored source.

**3. Stale comment.** Removed the old RAW-`driftFindings` half above `saveScanResult` so it tells one story now.

Also dropped a couple of Unreleased CHANGELOG notes, including the one-time "resolved" self-heal you flagged.

On the idea to push no closing line at all when signed in. I kept the signed-in line since it carries the "first deep scan free" CTA
But I'm happy to drop it and let the banner do the talking if you'd rather. Let me know and I'll adjust.

skhan75 (approved):
Noice!

PR #66, opened 2026-07-17
fix(security): skip comment lines in regex route fallback (#64 item 4)

### LINKED ISSUE #64: Four small honesty fixes in scan output (contributor-friendly batch)

BrandonGarate177:
## Summary

Addresses item 4 of #64 ("The suppression fallback can silently hide a route on a malformed line"). 

**Note:** The issue description was light on specifics for this item. Based on some investigation, I believe the scenario is: the regex fallback route extractors match route-like patterns inside comment lines, creating phantom `RouteInfo` entries that can steal `@vibedrift-public` annotations from adjacent real routes. If this isn't the intended scenario, happy to close this and adjust.

**The bug:**
```go
// @vibedrift-public                         ← annotation (line 6)
// r.POST("/admin/nuke", nukeHandler)        ← regex matches this comment (line 7)
r.POST("/admin/nuke", nukeHandler)           ← real route (line 8)
```

The regex creates a phantom route at line 7. The annotation on line 6 binds to the phantom (preceding-line match). The phantom gets suppressed and silently disappears from the vote. The real route on line 8 loses its intended suppression — it stays in the vote and gets flagged as "lacking auth" despite the user's `@vibedrift-public` intent.

**Fix:** Added comment-line skipping in all three regex fallback extractors (Go, JS/TS, Python) before matching route patterns:
- Go/JS: skip lines starting with `//` or `/*`
- Python: skip lines starting with `#` or `//`

The AST path was already immune (comment nodes never produce route registrations).

**Verified:** Existing test "route-shaped COMMENT" updated from asserting the old (buggy) behavior to the new behavior — both regex and AST paths now agree (no phantom routes from comments).

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Refactor
- [ ] Test
- [ ] Chore / infra

## Checklist

## Related issues

Partially closes #64 (item 4)

### DIFF
--- a/src/drift/security-consistency.ts
+++ b/src/drift/security-consistency.ts
@@ -340,6 +340,21 @@ function inheritedRateLimit(perRoute: boolean, fileMw: FileMiddleware | undefine
   return perRoute || (fileMw?.hasRateLimit ?? false);
 }
 
+// ─── Regex-fallback comment skipping ─────────────────────────────────
+// The regex route extractors (used when tree-sitter has no clean parse) match
+// route-shaped text line by line. A commented-out registration must NOT become
+// a phantom route — it would steal a @vibedrift-public annotation from the real
+// route below it (see #64 item 4). JS/TS and Go share C-style comments, so their
+// markers live in one place; Python differs (# line comments, """/''' docstrings).
+const C_STYLE_COMMENT_MARKERS = ["//", "/*"] as const; // JS, TS, Go
+const PYTHON_COMMENT_MARKERS = ["#"] as const;
+
+/** True when a source line is a line comment for the given markers. */
+function isCommentLine(line: string, markers: readonly string[]): boolean {
+  const trimmed = line.trimStart();
+  return markers.some((m) => trimmed.startsWith(m));
+}
+
 function extractGoRoutes(file: DriftFile, routes: RouteInfo[], fileMw: Map<string, FileMiddleware>, xfile: CrossFileIndex) {
   // AST only on a CLEAN parse: tree-sitter always returns a tree for broken Go
   // (with ERROR nodes), and error recovery SWALLOWS later valid registrations
@@ -368,9 +383,8 @@ function extractGoRoutesRegex(file: DriftFile, routes: RouteInfo[], fileMiddlewa
 
   for (let i = 0; i < lines.length; i++) {
     const line = lines[i];
-    const trimmed = line.trimStart();
-    // Skip comment lines — prevents phantom routes from commented-out code
-    if (trimmed.startsWith("//") || trimmed.startsWith("/*")) continue;
+    // Skip comment lines — prevents phantom routes from commented-out code.
+    if (isCommentLine(line, C_STYLE_COMMENT_MARKERS)) continue;
     let method = "", path = "";
     const echoMatch = line.match(echoPattern);
     if (echoMatch) { method = echoMatch[1]; path = echoMatch[2]; }
@@ -409,9 +423,8 @@ function extractJsRoutesRegex(file: DriftFile, routes: RouteInfo[], fileMiddlewa
   const expressPattern = /\.(?:get|post|put|patch|delete|all)\s*\(\s*['"`]([^'"`]+)['"`]/;
 
   for (let i = 0; i < lines.length; i++) {
-    const trimmed = lines[i].trimStart();
-    // Skip comment lines — prevents phantom routes from commented-out code
-    if (trimmed.startsWith("//") || trimmed.startsWith("/*")) continue;
+    // Skip comment lines — prevents phantom routes from commented-out code.
+    if (isCommentLine(lines[i], C_STYLE_COMMENT_MARKERS)) continue;
     const match = lines[i].match(expressPattern);
     if (!match) continue;
     const path = match[1];
@@ -506,10 +519,20 @@ function extractPythonRoutesRegex(file: DriftFile, routes: RouteInfo[], fileMidd
   const lines = file.content.split("\n");
   const routePattern = /@\w+\.(?:route|get|post|put|patch|delete)\s*\(\s*['"]([^'"]+)['"]/;
 
+  let inDocstring = false;
   for (let i = 0; i < lines.length; i++) {
-    const trimmed = lines[i].trimStart();
-    // Skip comment lines — prevents phantom routes from commented-out code
-    if (trimmed.startsWith("#") || trimmed.startsWith("//")) continue;
+    // Track triple-quoted docstring blocks: a route-shaped line inside a
+    // docstring is documentation, not a real registration. An odd count of
+    // triple-quotes on a line toggles the block state.
+    const tripleQuotes = (lines[i].match(/"""|'''/g) ?? []).length;
+    if (inDocstring) {
+      if (tripleQuotes % 2 === 1) inDocstring = false;
+      continue;
+    }
+    if (tripleQuotes % 2 === 1) { inDocstring = true; continue; }
+    // Skip comment lines — Python uses '#'. Prevents phantom routes from
+    // commented-out code.
+    if (isCommentLine(lines[i], PYTHON_COMMENT_MARKERS)) continue;
     const match = lines[i].match(routePattern);
     if (!match) continue;
     const path = match[1];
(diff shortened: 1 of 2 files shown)


src/drift/security-consistency.ts
skhan75, line 81: These kind of code additions will quickly become unreadable and unmanageable. We should add some property "type" to the file, so that one can easily do `file.type == PYTHON`, where RHS can be some enum and file now has a implicit type for file type. What do you think?

skhan75, line 204: okay we already have `file.language`. Let's make sure to reuse this property everywhere

skhan75, line 379: @BrandonGarate177  The code is becoming very unreadable, what do you think if we create separate classes per language? More code modularity, maintainability and easier to debug and diagnose. 

It's a total AI slop, which I am responsible for, but if you can try to organize few things a bit, that would be awesome.

src/drift/security-consistency.ts
```diff
@@ -406,6 +409,9 @@ function extractJsRoutesRegex(file: DriftFile, routes: RouteInfo[], fileMiddlewa
   const expressPattern = /\.(?:get|post|put|patch|delete|all)\s*\(\s*['"`]([^'"`]+)['"`]/;
 
   for (let i = 0; i < lines.length; i++) {
+    const trimmed = lines[i].trimStart();
+    // Skip comment lines — prevents phantom routes from commented-out code
+    if (trimmed.startsWith("//") || trimmed.startsWith("/*")) continue;
```
skhan75: JS and Go share the same regex? Why not add reusability? Think about Enums and Constants, and see if it could help make things a little more cleaner here

src/drift/security-consistency.ts
```diff
@@ -501,6 +507,9 @@ function extractPythonRoutesRegex(file: DriftFile, routes: RouteInfo[], fileMidd
   const routePattern = /@\w+\.(?:route|get|post|put|patch|delete)\s*\(\s*['"]([^'"]+)['"]/;
 
   for (let i = 0; i < lines.length; i++) {
+    const trimmed = lines[i].trimStart();
+    // Skip comment lines — prevents phantom routes from commented-out code
+    if (trimmed.startsWith("#") || trimmed.startsWith("//")) continue;
```
skhan75: Python doesn't have `//`. There is multi line `"""` though, did you mean that?

src/drift/security-consistency.ts
skhan75, line 137: I know this is beyond the context of your PR, but since I have comments around refactor, one thing we can bring in, is to start using typescript interfaces. I just hate how AI slop everything around, though it works, but the code is just so cluttered. 
This is a good one if you want to take a dig at it.

skhan75 (changes requested):
Noice work! @BrandonGarate177 
I have few comments around refactoring some things. But overall solid progress.

src/drift/security-consistency.ts
BrandonGarate177: fixed to `#` only, and added `"""`/`'''` docstring tracking so a route inside a docstring doesn't get grabbed. Test added.

src/drift/security-consistency.ts
BrandonGarate177: I'm deferring this for #69, I'm leaning towards reusing the existing `SupportedLanguage` instead of adding a new enum tho

skhan75: Handled in #69

skhan75 (approved):

PR #67, opened 2026-07-17
fix(deps): resolve Go imports per nested module (#48)

### LINKED ISSUE #48: [bug]  Go dependency analyzer reports false "missing" imports from nested modules
## What happened

Running VibeDrift on Go repos with nested `go.mod` files (multi-module repos) produces false "imports not in go.mod" findings. The analyzer reads only the root `go.mod` but scans all `.go` files in the tree, including subdirectories that have their own `go.mod` with their own dependency declarations.

This affects **production code**, not just examples — HashiCorp Terraform has 11 nested modules in `internal/` and produces 38 false positives.

### Case 1: go-chi/chi (2 false positives)

```bash
npx @vibedrift/cli --local-only /path/to/chi
```

**Output:**
```
── Dependency Health ───────────────────────────────
  ✘ 2 Go imports not in go.mod: github.com/go-chi/docgen, github.com/go-chi/render
    go.mod
```

**Proof these ARE declared (in the nested module):**
```
chi/_examples/rest/go.mod:7:  github.com/go-chi/docgen v1.2.0
chi/_examples/rest/go.mod:8:  github.com/go-chi/render v1.0.1
```

**Where they're imported:**
```
_examples/rest/main.go:49:  "github.com/go-chi/docgen"
_examples/rest/main.go:50:  "github.com/go-chi/render"
```

### Case 2: HashiCorp Terraform (38 false positives — production code)

```bash
npx @vibedrift/cli --local-only /path/to/terraform
```

**Output:**
```
── Dependency Health ───────────────────────────────
  ⚠ 232 potentially unused Go modules: ...
    go.mod
  ✘ 38 Go imports not in go.mod: go.opentelemetry.io/otel/codes, github.com/hashicorp/go-azure-sdk,
    github.com/hashicorp/consul...
    go.mod
```

**Proof these ARE declared (in their respective nested modules):**
```
internal/backend/remote-state/azure/go.mod:7:  github.com/hashicorp/go-azure-sdk/resource-manager v0.20250131.1134653
internal/backend/remote-state/azure/go.mod:8:  github.com/hashicorp/go-azure-sdk/sdk v0.20250131.1134653
internal/backend/remote-state/consul/go.mod:6: github.com/hashicorp/consul/api v1.32.1
internal/backend/remote-state/consul/go.mod:7: github.com/hashicorp/consul/sdk v0.16.1
```

**Where they're imported (production source code):**
```
internal/backend/remote-state/azure/api_client.go:15: "github.com/hashicorp/go-azure-sdk/resource-manager/storage/..."
internal/backend/remote-state/azure/api_client.go:16: "github.com/hashicorp/go-azure-sdk/sdk/auth"
internal/backend/remote-state/consul/backend.go:11:   consulapi "github.com/hashicorp/consul/api"
internal/backend/remote-state/consul/client.go:19:    consulapi "github.com/hashicorp/consul/api"
```

## Expected behavior

Go files under a directory with its own `go.mod` should be checked against that module's dependencies, not the root module's. Alternatively, they should be excluded from the root module's dependency check entirely.

## Environment

- VibeDrift version: 0.15.0
- Node version: v24.12.0
- OS: Windows 11 (platform-independent issue)

### Terraform's nested module layout (all production code):
```
terraform/
├── go.mod                                          ← root module (VibeDrift reads only this)
├── internal/backend/remote-state/azure/go.mod      ← declares go-azure-sdk
├── internal/backend/remote-state/consul/go.mod     ← declares consul
├── internal/backend/remote-state/cos/go.mod        ← declares cos SDK
├── internal/backend/remote-state/gcs/go.mod        ← declares GCS SDK
├── internal/backend/remote-state/kubernetes/go.mod ← declares k8s client
├── internal/backend/remote-state/oci/go.mod        ← declares OCI SDK
├── internal/backend/remote-state/oss/go.mod        ← declares OSS SDK
├── internal/backend/remote-state/pg/go.mod         ← declares pq driver
├── internal/backend/remote-state/s3/go.mod         ← declares AWS SDK
└── internal/legacy/go.mod                          ← legacy module
```

This is standard Go multi-module architecture for large projects. Not just examples — these are production backends for Terraform's state storage.

### Who's affected:
- Any Go repo with `tools/go.mod` (common for build tools)
- Any Go workspace (`go.work`) with multiple modules
- Monorepos with nested services (`services/auth/go.mod`, `services/api/go.mod`)
- Major projects: Terraform, CockroachDB, Vitess, etc.

skhan75:
Fixes #48.

## What was wrong

On Go repos with more than one `go.mod`, the dependency scan read only the root `go.mod` but checked every `.go` file in the tree against it. Files that live under a nested module (an example module, a `tools/` module, a `go.work` service, a nested backend) got their imports compared against the wrong dependency list, so packages that are plainly declared, just in a different module, were reported as missing.

Brandon measured this on real repos: 2 false errors on go-chi/chi and 38 on HashiCorp Terraform, all production or example code where the import really is declared in a nested `go.mod`.

## Before and after (go-chi/chi)

The two imports below are declared in `_examples/rest/go.mod`, not the root, so the old scan called them missing.

**Before** — `@vibedrift/cli` 0.16.2

before

**After** — this PR

after

Dependency Health goes fully clean on chi. The Vibe Drift Score is unchanged at 97/100, which is the point: this removes a false alarm without moving the number.

## What changed

- **Per-module resolution.** Every nested `go.mod` under the scan root is parsed, and each `.go` file is checked against its *nearest enclosing* module rather than the root one.
- **Prefix-aware module matching.** A declared module now matches an import when it is the import path or a path-segment prefix of it. The old code truncated every import to three segments, which broke multi-segment module paths like `github.com/hashicorp/go-azure-sdk/sdk` and `/vN` major-version suffixes.
- **Sibling and indirect handling.** An import that resolves to another in-repo module is never called missing, and `// indirect` requires are excluded from the unused-module check since they are transitive and never imported directly.
- **No new false positives.** The nested-`go.mod` walk does not apply the main scan's gitignore or file-count cap, so a nested module can be discovered while its files are not in the scan. Those modules are no longer phantom-checked (no scanned files means no evidence of what they import), and files under a nested `go.mod` we could not parse are excluded rather than blamed on root. Both paths are covered by tests.

Single-module repos are unchanged.

## Tests

`test/unit/analyzers/dependencies.test.ts` and `test/unit/core/gomod.test.ts` cover: the chi nested-example case, Terraform-style multi-segment module paths, `/vN` suffixes, sibling in-repo imports, `// indirect` requires, per-module phantom accounting, the gitignored/capped empty-module case, the unparseable-nested-module case, and single-module regression. Full suite green (1869 tests), lint clean.

Scoring version moves to v13 (Go dependency drift is what changed), which the existing silent-migration path absorbs.


src/analyzers/dependencies.ts
```diff
@@ -280,50 +283,138 @@ function analyzeJsDeps(ctx: AnalysisContext): Finding[] {
...
+  // service modules): each .go file is checked against its NEAREST enclosing
+  // go.mod, not the root one. Nested modules declare their own dependencies.
+  // `directPaths` drives phantom detection: `// indirect` requires are
+  // transitive deps go mod tidy records, never imported directly, so flagging
+  // them "unused" would be a false positive. `allPaths` (direct + indirect)
+  // drives the missing-import check — an import satisfied by an indirect entry
+  // is still declared.
+  const scopes = [
+    {
+      dir: "",
```
skhan75, line 357: Todo: Create scope interfaces

src/analyzers/dependencies.ts
```diff
@@ -280,50 +283,138 @@ function analyzeJsDeps(ctx: AnalysisContext): Finding[] {
...
+  // are never "missing" — the code is physically present, and whether a
+  // workspace declares the cross-module require is build tooling, not drift.
+  const inRepoModules = scopes.map((s) => s.module).filter((m) => m.length > 0);
 
+  const goFiles = ctx.files.filter((f) => f.language === "go");
   for (const file of goFiles) {
-    const goImports = extractGoImports(file.content);
-    for (const importPath of goImports) {
+    const rel = file.relativePath.replace(/\\/g, "/");
+    const scope = byDepth.find((s) => s.dir === "" || rel.startsWith(s.dir + "/"))!;
```
skhan75: For uncle Claude --> heard of Trie before? We already store import graph ast information? Why not build a trie to store mod file as boolean `hasModPackage` at every directory? Or correct me if I am wrong? I just don't the  current way you implemented is a good engineering design and scalable?

src/analyzers/dependencies.ts
skhan75: Done in 3a7eea0. Pulled the per-module scope object out into a named `ModuleScope` interface (dir, module, goModPath, directPaths, allPaths, imports, fileCount), so the two inline literals now share one documented shape instead of being duplicated by hand.

🤖 Reply drafted with [Claude Code](https://claude.com/claude-code)

src/analyzers/dependencies.ts
skhan75: Good call on the direction, took it in 3a7eea0. One honest note on the framing first, then what I built.

The old `byDepth.find` was linear in the number of **modules** (go.mod files), not directories and not files squared. For the repo in this issue, Terraform, that is 11 modules, so per file it was roughly 11 cheap `startsWith` calls, about 55k comparisons across the whole tree. It was not actually a bottleneck. Where your idea genuinely wins is readability and the pathological case: a monorepo with thousands of go.mod files, where `O(files x modules)` would start to bite.

So I implemented it as a sparse `Map<dir, ModuleScope>` keyed only by real module directories, plus an upward ancestor walk from each file's directory (`nearestModuleScope`). That is `O(path depth)` per file with O(1) lookups, independent of how many modules the repo has. I went with the sparse map rather than a dense trie carrying a boolean at every directory, because the map gives the same `O(depth)` lookup while using `O(modules)` memory instead of `O(all directories)`, which matters on a big tree.

One place I did not follow the suggestion: I did not reuse the import graph AST for this. That index models file to imported-module edges (who imports what), whereas this needs directory to containing-module (which go.mod owns a subtree). Different question, different shape, so a small dedicated map is a cleaner fit than overloading the import graph with a second meaning.

Added two tests for the tricky part the new walk introduces: a module nested inside another module (deepest ancestor wins) and a parsed module nested inside an opaque one.

🤖 Reply drafted with [Claude Code](https://claude.com/claude-code)

Issue #56, opened 2026-07-16
Import-style / convention drift is JS/TS only; add Python, Go, Rust


skhan75:
## Gap

VibeDrift analyzes **import style and convention drift for JavaScript and TypeScript only**. There is no equivalent check for Python, Go, or Rust, so a repo that mixes import conventions in those languages gets no signal.

## Current state (verified against source)

- `src/analyzers/imports.ts` — ESM vs CJS (`require` vs `import`), gated `applicableLanguages: ["javascript", "typescript"]`.
- `src/drift/import-consistency.ts` — relative vs alias import paths, gated to JS/TS (line ~29).
- `src/core/import-graph-ast.ts` — the AST import graph parses JS/TS import syntax and feeds `dead-code`, `phantom-scaffolding`, and `dependencies`.
- `src/analyzers/dependencies.ts` — phantom/missing dependency detection for JS/TS (via `package.json`) and Go (via `go.mod`).

So imports in other languages are **parsed for dependency and dead-code purposes** (Go via `go.mod`), but never analyzed as a **style/convention drift** signal.

## What's missing per language

- **Python**: no absolute-vs-relative convention, import ordering, or `from x import *` check.
- **Go**: no import grouping (stdlib / third-party / local) or ordering convention.
- **Rust**: no `use` convention (grouping, glob `use ...::*`, `self`/`super`/`crate` style).

## Proposed direction

Either add a per-language import-style analyzer, or generalize the existing `import-consistency` dominance-vote to Go/Python/Rust once the AST import graph extracts their import nodes. Reuse the shared drift machinery: directory-scoped dominance vote, the entropy gate (stay silent when there is no dominant convention), and intent-hint seeding.

## Acceptance criteria

- A directory-scoped import-style drift check for at least one new language, wired into the drift registry with a weight.
- Calibration fixtures showing it fires on a real mixed-convention directory and stays silent when a directory has no dominant convention.
- No regression to JS/TS import behavior.

## Context

Surfaced while writing the Developer Handbook; the analyzer summary table now documents this gap explicitly. Error handling, by contrast, *is* covered for Go/Python/Rust (via the `language-specific` analyzer), so this issue is scoped to imports only.

---
*Filed via Claude Code on behalf of the maintainer.*

PR #70, opened 2026-07-20
refactor(security): per-language route-extractor modules behind Route…

### LINKED ISSUE #69: [feat] Modularize security route extraction into per-language functional modules
## Problem / motivation

While reviewing #66 (the item-4 comment-skip fix for #64), @skhan75 flagged maintainability concerns in `src/drift/security-consistency.ts`: it dispatches route extraction by branching on `file.language` inline, concentrates per-language regex fallbacks in one growing file, and re-derives language from the path extension (`langOfFile`) rather than reusing what the pipeline already computed. The file is hard to read and risky to change — which matters because it's a security-critical detector.

Capturing @skhan75's review comments here so they live in one durable place (the inline threads on #66 go stale once that PR is revised). This is his suggestion, filed to make it actionable.

Original review threads on #66:
- Separate units per language: https://github.com/VibeDrift/VibeDrift/pull/66#discussion_r3604789932
- File-type / reuse `file.language`: https://github.com/VibeDrift/VibeDrift/pull/66#discussion_r3604770980 and https://github.com/VibeDrift/VibeDrift/pull/66#discussion_r3604762873
- TypeScript interfaces: https://github.com/VibeDrift/VibeDrift/pull/66#discussion_r3604829821
- Shared marker constants: https://github.com/VibeDrift/VibeDrift/pull/66#discussion_r3604799570 *(already addressed in #66)*

### Review notes from #66 this issue resolves

The following review notes on #66 are deferred here (that PR stays scoped to the item-4 bug fix):

- [ ] Separate per-language units — implemented here as **functional modules** behind a shared `RouteExtractor` interface
- [ ] File-`type` / reuse `file.language` in place of `langOfFile` (extension re-derivation)
- [ ] Named TypeScript interfaces for the extractor contract
- [x] Shared comment-marker constants — *already done in #66*

## Proposed solution

Give each language a self-contained **functional module** (matching the existing `security-ast-*.ts` style — a stateless `(file) → RouteInfo[]` transform doesn't warrant a class) exporting an object that satisfies a shared `RouteExtractor` interface. `security-consistency.ts` becomes thin dispatch over a `Record<SupportedLanguage, RouteExtractor>` table instead of inline `if (file.language === ...)` branches.

```ts
interface RouteExtractor {
  language: SupportedLanguage;
  // Runs the AST-clean-parse check internally and falls back to regex,
  // delegating AST work to the existing security-ast-<lang>.ts function.
  extract(file: DriftFile, deps: ExtractDeps): RouteInfo[];
}
```

Reuse the existing `SupportedLanguage` (`src/core/language.ts`) as the dispatch key — no new `FileType` enum.

**In scope:** the per-language route extractors (`extract<Lang>Routes` wrappers + their regex fallbacks) and the dispatch in `security-consistency.ts`; named interfaces for the extractor contract.

**Out of scope:** the AST module internals (`security-ast-*.ts`, delegated to rather than rewritten), suppression, finding construction, and any change to the `RouteInfo` shape or its baseline serialization. (`langOfFile` is removed only if doable without adding a field to the serialized `RouteInfo`; otherwise it's a follow-up.)

## How does this improve drift measurement?

Neutral infrastructure (maintainability), with an accuracy angle: `security-consistency.ts` is where the recent honesty/correctness fixes for #64 landed, and its inline-branching sprawl is exactly where subtle per-language extraction bugs hide (the phantom-route bug in #66 was one). A per-language interface with isolated modules reduces the surface for those inconsistencies and makes adding route-style support for more languages (cf. #56) safer — so it serves *more confident* measurement even though it changes no scoring today.

## Alternatives considered

- **Classes per language** (as literally phrased in review): rejected — the extractors are stateless, so classes add construction ceremony without state to justify it and diverge from the codebase's functional style. Factory functions would cover any future DI need.
- **New `FileType` enum:** rejected — duplicates the existing `SupportedLanguage`.
- **Refactor the whole security-drift surface** (AST modules, suppression, finding construction): rejected — too large a diff on security-critical code; start with the extractors and extend the pattern later.
- **Thread language onto `RouteInfo`:** rejected — `RouteInfo` is serialized byte-identically into the baseline; adding a field changes the persisted format.

## Additional context

**Current state (verified):** `security-consistency.ts` branches on `file.language`, then calls per-language wrappers that do an AST-clean check and fall back to a local `extract<Lang>RoutesRegex`. AST extractors already live in `security-ast-{go,python,rust}.ts`. `RouteInfo` is serialized byte-identically into the baseline (see the `authUnsureHook` doc-comment).

**Acceptance criteria:**
- Behavior-preserving: no change to which routes are extracted, suppressed, or flagged; pinned-legacy regex tests and `test/unit/drift/security-*.test.ts` pass unchanged.
- `RouteInfo` baseline serialization stays byte-identical.
- Route dispatch is a `SupportedLanguage`-keyed table; per-language extraction lives in dedicated functional modules behind `RouteExtractor`.
- `npm run lint`, `npm run typecheck`, `npm test`, `npm run build` all pass.

**Related:** surfaced in review of #66 (#64 item 4); the shared marker-constants piece is already handled in #66. Complements #56 (multi-language import-style).

BrandonGarate177:
## Summary

Modularizes security route extraction out of `src/drift/security-consistency.ts`, addressing the refactor discussion from #66 (tracked in #69). The file previously branched on `file.language` inline and concentrated every language's regex fallback in one place.

- New `src/drift/route-extractors/`:
  - `types.ts` — `RouteInfo`, `FileMiddleware`, and the `ExtractDeps` / `RouteExtractor` interfaces.
  - `shared.ts` — comment-marker constants, `isCommentLine`, `inheritedAuth/Validation/RateLimit`, `findHandlerContent`, `MUTATION_METHODS`.
  - `patterns.ts` — every per-language regex as a named constant (route patterns with their capture groups, plus the auth/validation/rate-limit/error-handler detectors), so they're individually testable.
  - `go.ts`, `js.ts`, `python.ts`, `rust.ts` — each exports a functional `RouteExtractor` (AST on a clean parse, regex fallback otherwise). `js.ts` serves both the `javascript` and `typescript` keys; `rust.ts` is AST-only.
- `security-consistency.ts` is now thin dispatch over a total `Record<SupportedLanguage, RouteExtractor>` table instead of inline `if/else` — a missing language is a compile error at the literal, with a runtime `in` guard for unknown/unset `file.language`. It re-exports `RouteInfo`/`FileMiddleware` so `security-ast{,-go,-python,-rust}.ts` and `security-suppression.ts` are **untouched**.
- Reuses the pipeline's `file.language` and removes the `langOfFile` extension re-derivation.

**Design note:** per the plan approved on #69, the per-language units are **functional modules** behind a shared interface (not classes) — the extractors are stateless `(file) → RouteInfo[]` transforms, matching the existing `security-ast-*` style. `SupportedLanguage` is reused as the dispatch key (no new `FileType` enum), and `RouteExtractor` carries no `language` field — dispatch is by the table key, the single source of truth.

**Out of scope (deliberately):** the AST module internals, suppression, finding construction, and any change to the `RouteInfo` shape or its baseline serialization.

## Tests

Added in response to review — the refactor moved code into new files, so these pin the behavior in its new home:

- `test/unit/drift/route-extractors.test.ts` — calls each extractor directly on tree-less files and asserts the **full `RouteInfo` via `toEqual`**: comment-skip (JS/Go C-style, Python `#` + docstring), auth/validation/rate-limit detection, `inheritedValidation`/`inheritedRateLimit` via file-level middleware, error-handler, and method/path capture (Echo vs Gorilla `.Methods`, `methods=[...]`, `.all`).
- `test/unit/drift/patterns.test.ts` — per-language regex units: route patterns' capture groups + each detector's match/reject.

**Guardrail check:** I reproduced the reviewer's experiment — deleting a comment-skip line (or gutting `inheritedValidation`/`inheritedRateLimit`) now turns `route-extractors.test.ts` **red**, closing the gap where those breakages previously went unnoticed.

## Verification

- The refactor is **behavior-preserving**: regexes were relocated verbatim, and the `Record`/`language`-field changes are type-only / dead-code. AST modules, suppression, finding construction, and `RouteInfo` serialization are untouched. (Route output was independently confirmed byte-identical during review — 51 routes across all four languages + three real repos.)
- Full suite: **1906 passed (+27 new tests)**. The only failures are two pre-existing, environment-only tests unrelated to this diff — `git-metadata` (git-timing flakiness on Windows) and `concise-ai-summary` (an ANSI/`FORCE_COLOR` assertion); neither file is touched here and both are green on CI.
- `npm run typecheck` clean, `npm run build` success, `npm run lint` 0 errors (no issues in the new modules).

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Documentation
- [x] Refactor
- [x] Test
- [ ] Chore / infra

## Checklist

- [x] New behavior has tests — *added `route-extractors.test.ts` (full-`RouteInfo` boundary tests) and `patterns.test.ts` (regex units); verified they fail under a deliberate comment-skip deletion.*
- [x] `README.md` updated if a flag, command, or feature changed — *N/A: no user-facing change.*
- [x] This change improves how accurately or confidently VibeDrift measures drift (or is neutral infra) — *neutral infra: reduces the surface where per-language extraction bugs hide (cf. the #66 phantom-route bug), now with test coverage, and makes adding new languages (cf. #56) safer.*

## Related issues

Closes #69

### DIFF
--- a/src/drift/route-extractors/go.ts
+++ b/src/drift/route-extractors/go.ts
@@ -15,44 +15,49 @@ import {
   inheritedRateLimit,
   findHandlerContent,
 } from "./shared.js";
+import {
+  GO_ROUTE_ECHO,
+  GO_ROUTE_GORILLA,
+  GO_AUTH,
+  GO_VALIDATION,
+  GO_RATE_LIMIT,
+  GO_ERROR_HANDLER,
+} from "./patterns.js";
 
 function extractGoRoutesRegex(file: DriftFile, fileMiddleware: FileMiddleware | undefined): RouteInfo[] {
   const routes: RouteInfo[] = [];
   const lines = file.content.split("\n");
-  const echoPattern = /\.\b(GET|POST|PUT|PATCH|DELETE|HEAD|OPTIONS)\s*\(\s*"([^"]+)"/;
-  const gorillaPattern = /HandleFunc\s*\(\s*"([^"]+)".*\.Methods\s*\(\s*"(\w+)"/;
 
   for (let i = 0; i < lines.length; i++) {
     const line = lines[i];
     // Skip comment lines — prevents phantom routes from commented-out code.
     if (isCommentLine(line, C_STYLE_COMMENT_MARKERS)) continue;
     let method = "", path = "";
-    const echoMatch = line.match(echoPattern);
+    const echoMatch = line.match(GO_ROUTE_ECHO);
     if (echoMatch) { method = echoMatch[1]; path = echoMatch[2]; }
-    const gorillaMatch = line.match(gorillaPattern);
+    const gorillaMatch = line.match(GO_ROUTE_GORILLA);
     if (gorillaMatch) { path = gorillaMatch[1]; method = gorillaMatch[2]; }
     if (!method || !path) continue;
 
     const context = lines.slice(Math.max(0, i - 10), i + 10).join("\n");
     const handlerContent = findHandlerContent(file.content, path);
 
-    const perAuth = /[Aa]uth|[Tt]oken|require[A-Z]|middleware\.\w*[Aa]uth/.test(context);
-    const perVal = /[Bb]ind|[Vv]alidat|[Pp]arse/.test(handlerContent);
-    const perRate = /[Rr]ate[Ll]imit|[Tt]hrottle/.test(context + handlerContent);
+    const perAuth = GO_AUTH.test(context);
+    const perVal = GO_VALIDATION.test(handlerContent);
+    const perRate = GO_RATE_LIMIT.test(context + handlerContent);
 
     routes.push({
       method, path, file: file.relativePath, line: i + 1,
       hasAuth: inheritedAuth(perAuth, fileMiddleware),
       hasValidation: inheritedValidation(perVal, fileMiddleware),
       hasRateLimit: inheritedRateLimit(perRate, fileMiddleware),
-      hasErrorHandler: /catch|err\s*!=\s*nil|try|except|\.catch/.test(handlerContent),
+      hasErrorHandler: GO_ERROR_HANDLER.test(handlerContent),
     });
   }
   return routes;
 }
 
 export const goRouteExtractor: RouteExtractor = {
-  language: "go",
   extract(file, deps) {
     // AST only on a CLEAN parse: tree-sitter always returns a tree for broken Go
     // (with ERROR nodes), and error recovery SWALLOWS later valid registrations
--- a/src/drift/route-extractors/go.ts
+++ b/src/drift/route-extractors/go.ts
@@ -1,0 +1,71 @@
+/**
+ * Go route extractor — Echo / Gin / Gorilla mux.
+ *
+ * AST on a CLEAN parse (delegated to security-ast-go), regex fallback otherwise.
+ */
+
+import type { DriftFile } from "../types.js";
+import { extractGoRoutesAst } from "../security-ast-go.js";
+import type { RouteInfo, FileMiddleware, RouteExtractor } from "./types.js";
+import {
+  C_STYLE_COMMENT_MARKERS,
+  isCommentLine,
+  inheritedAuth,
+  inheritedValidation,
+  inheritedRateLimit,
+  findHandlerContent,
+} from "./shared.js";
+
+function extractGoRoutesRegex(file: DriftFile, fileMiddleware: FileMiddleware | undefined): RouteInfo[] {
+  const routes: RouteInfo[] = [];
+  const lines = file.content.split("\n");
+  const echoPattern = /\.\b(GET|POST|PUT|PATCH|DELETE|HEAD|OPTIONS)\s*\(\s*"([^"]+)"/;
+  const gorillaPattern = /HandleFunc\s*\(\s*"([^"]+)".*\.Methods\s*\(\s*"(\w+)"/;
+
+  for (let i = 0; i < lines.length; i++) {
+    const line = lines[i];
+    // Skip comment lines — prevents phantom routes from commented-out code.
+    if (isCommentLine(line, C_STYLE_COMMENT_MARKERS)) continue;
+    let method = "", path = "";
+    const echoMatch = line.match(echoPattern);
+    if (echoMatch) { method = echoMatch[1]; path = echoMatch[2]; }
+    const gorillaMatch = line.match(gorillaPattern);
+    if (gorillaMatch) { path = gorillaMatch[1]; method = gorillaMatch[2]; }
+    if (!method || !path) continue;
+
+    const context = lines.slice(Math.max(0, i - 10), i + 10).join("\n");
+    const handlerContent = findHandlerContent(file.content, path);
+
+    const perAuth = /[Aa]uth|[Tt]oken|require[A-Z]|middleware\.\w*[Aa]uth/.test(context);
+    const perVal = /[Bb]ind|[Vv]alidat|[Pp]arse/.test(handlerContent);
+    const perRate = /[Rr]ate[Ll]imit|[Tt]hrottle/.test(context + handlerContent);
+
+    routes.push({
+      method, path, file: file.relativePath, line: i + 1,
+      hasAuth: inheritedAuth(perAuth, fileMiddleware),
+      hasValidation: inheritedValidation(perVal, fileMiddleware),
+      hasRateLimit: inheritedRateLimit(perRate, fileMiddleware),
+      hasErrorHandler: /catch|err\s*!=\s*nil|try|except|\.catch/.test(handlerContent),
+    });
+  }
+  return routes;
+}
+
+export const goRouteExtractor: RouteExtractor = {
+  language: "go",
+  extract(file, deps) {
+    // AST only on a CLEAN parse: tree-sitter always returns a tree for broken Go
+    // (with ERROR nodes), and error recovery SWALLOWS later valid registrations
+    // into a broken call's argument_list as clean-looking nested calls (a
+    // cross-bless hazard). Any parse error routes the whole file to the regex,
+    // byte-identical to today's behavior INCLUDING the regex path's known
+    // over-blesses (see the pinned-legacy tests).
+    if (file.tree && !file.tree.rootNode.hasError) {
+      // The cross-file index lets an imported package middleware selector resolve
+      // to its in-repo defining body; an unresolved / external / index-disabled
+      // selector stays UNSURE, byte-identical to the in-file-only path.
+      return extractGoRoutesAst(file.tree, file.relativePath, deps.xfile);
+    }
+    return extractGoRoutesRegex(file, deps.fileMw.get(file.relativePath));
+  },
+};
--- a/src/drift/route-extractors/js.ts
+++ b/src/drift/route-extractors/js.ts
@@ -16,38 +16,44 @@ import {
   inheritedValidation,
   inheritedRateLimit,
 } from "./shared.js";
+import {
+  JS_ROUTE,
+  JS_METHOD,
+  JS_AUTH,
+  JS_VALIDATION,
+  JS_RATE_LIMIT,
+  JS_ERROR_HANDLER,
+} from "./patterns.js";
 
 function extractJsRoutesRegex(file: DriftFile, fileMiddleware: FileMiddleware | undefined): RouteInfo[] {
   const routes: RouteInfo[] = [];
   const lines = file.content.split("\n");
-  const expressPattern = /\.(?:get|post|put|patch|delete|all)\s*\(\s*['"`]([^'"`]+)['"`]/;
 
   for (let i = 0; i < lines.length; i++) {
     // Skip comment lines — prevents phantom routes from commented-out code.
     if (isCommentLine(lines[i], C_STYLE_COMMENT_MARKERS)) continue;
-    const match = lines[i].match(expressPattern);
+    const match = lines[i].match(JS_ROUTE);
     if (!match) continue;
     const path = match[1];
-    const method = match[0].match(/\.(get|post|put|patch|delete|all)/)?.[1]?.toUpperCase() ?? "ANY";
+    const method = match[0].match(JS_METHOD)?.[1]?.toUpperCase() ?? "ANY";
     const context = lines.slice(Math.max(0, i - 5), i + 20).join("\n");
 
-    const perAuth = /(?:requireAuth|isAuthenticated|passport\.authenticate|verifyToken|jwt|authMiddleware)/.test(context);
-    const perVal = /validate|joi|zod|yup|celebrate|body\(|query\(/.test(context);
-    const perRate = /rateLimit|throttle|limiter/.test(context);
+    const perAuth = JS_AUTH.test(context);
+    const perVal = JS_VALIDATION.test(context);
+    const perRate = JS_RATE_LIMIT.test(context);
 
     routes.push({
       method, path, file: file.relativePath, line: i + 1,
       hasAuth: inheritedAuth(perAuth, fileMiddleware),
       hasValidation: inheritedValidation(perVal, fileMiddleware),
       hasRateLimit: inheritedRateLimit(perRate, fileMiddleware),
-      hasErrorHandler: /catch|try|\.catch|next\(err/.test(context),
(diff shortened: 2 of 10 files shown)


skhan75:
I see few test failures in the description above. Can you confirm?

src/drift/route-extractors/types.ts
```diff
@@ -0,0 +1,61 @@
...
+import type { SupportedLanguage } from "../../core/types.js";
+import type { CrossFileIndex } from "../security-xfile-index.js";
+
+export interface FileMiddleware {
+  hasAuth: boolean;
+  hasValidation: boolean;
+  hasRateLimit: boolean;
+}
+
+export interface RouteInfo {
```
skhan75, line 20: 🔥

src/drift/route-extractors/go.ts
```diff
@@ -0,0 +1,71 @@
...
+    const perAuth = /[Aa]uth|[Tt]oken|require[A-Z]|middleware\.\w*[Aa]uth/.test(context);
+    const perVal = /[Bb]ind|[Vv]alidat|[Pp]arse/.test(handlerContent);
+    const perRate = /[Rr]ate[Ll]imit|[Tt]hrottle/.test(context + handlerContent);
+
+    routes.push({
+      method, path, file: file.relativePath, line: i + 1,
+      hasAuth: inheritedAuth(perAuth, fileMiddleware),
+      hasValidation: inheritedValidation(perVal, fileMiddleware),
+      hasRateLimit: inheritedRateLimit(perRate, fileMiddleware),
+      hasErrorHandler: /catch|err\s*!=\s*nil|try|except|\.catch/.test(handlerContent),
```
skhan75: I think we need to put error handler regexes in a separate file and test them individually to make sure for every language the regex captures the correct match group

src/drift/route-extractors/js.ts
```diff
@@ -0,0 +1,58 @@
...
+    const perAuth = /(?:requireAuth|isAuthenticated|passport\.authenticate|verifyToken|jwt|authMiddleware)/.test(context);
+    const perVal = /validate|joi|zod|yup|celebrate|body\(|query\(/.test(context);
+    const perRate = /rateLimit|throttle|limiter/.test(context);
+
+    routes.push({
+      method, path, file: file.relativePath, line: i + 1,
+      hasAuth: inheritedAuth(perAuth, fileMiddleware),
+      hasValidation: inheritedValidation(perVal, fileMiddleware),
+      hasRateLimit: inheritedRateLimit(perRate, fileMiddleware),
+      hasErrorHandler: /catch|try|\.catch|next\(err/.test(context),
```
skhan75: same here

src/drift/route-extractors/python.ts
```diff
@@ -0,0 +1,136 @@
...
+    const perAuth = /login_required|jwt_required|@requires|permission|token/.test(context);
+    const perVal = /pydantic|validate|Schema|Serializer/.test(context);
+    const perRate = /rate_limit|throttle|limiter/.test(context);
+
+    routes.push({
+      method, path, file: file.relativePath, line: i + 1,
+      hasAuth: inheritedAuth(perAuth, fileMiddleware),
+      hasValidation: inheritedValidation(perVal, fileMiddleware),
+      hasRateLimit: inheritedRateLimit(perRate, fileMiddleware),
+      hasErrorHandler: /try|except|raise/.test(context),
```
skhan75: same here

skhan75 (changes requested):

src/drift/security-consistency.ts
```diff
@@ -299,280 +274,34 @@ function buildFileMiddlewareIndex(files: DriftFile[]): Map<string, FileMiddlewar
 
 // ─── Route extraction (per-route middleware via proximity) ──────────
 
+// Route extraction dispatched by language. Each per-language module is a
+// self-contained functional RouteExtractor (AST on a clean parse, regex
+// fallback otherwise). JS and TS share one extractor.
+const ROUTE_EXTRACTORS: Partial<Record<SupportedLanguage, RouteExtractor>> = {
```
skhan75: Small one: drop `Partial` here.

The table already lists all five languages, and plain `Record<SupportedLanguage, RouteExtractor>` compiles fine today (I checked, `tsc --noEmit` exits 0).

Why it matters: with `Partial`, if someone adds a 6th language later and forgets to add it here, nothing complains. The lookup returns `undefined`, the file gets skipped, and we silently report **zero routes** for that language. Without `Partial`, it's a compile error at this literal, which is the whole point of having a table.

(Also, the PR description says `Record` but the code says `Partial<Record>`.)

src/drift/route-extractors/types.ts
```diff
@@ -0,0 +1,61 @@
...
+  xfile: CrossFileIndex;
+}
+
+/**
+ * A per-language route extractor. `extract` runs the AST path on a clean parse
+ * and falls back to regex otherwise, returning the routes for a single file.
+ * Stateless — a plain object satisfies it (no class needed).
+ */
+export interface RouteExtractor {
+  language: SupportedLanguage;
```
skhan75: This field is never read anywhere. Dispatch uses the table key (`file.language` at `security-consistency.ts:296`), never `extractor.language`.

It's also already inconsistent: `jsRouteExtractor` says `language: "javascript"` but it's registered under **both** `javascript` and `typescript`.

The trap is if someone later does the obvious thing:

```ts
Object.fromEntries(EXTRACTORS.map(e => [e.language, e]))
```

TypeScript route extraction silently disappears. No type error, no failing test.

Either delete the field, or make it `languages: SupportedLanguage[]` so JS can honestly say `["javascript", "typescript"]`.

To be fair this came straight from the interface sketch in #69, so it's my spec bug, not something you introduced.

src/drift/route-extractors/go.ts
```diff
@@ -0,0 +1,71 @@
...
+function extractGoRoutesRegex(file: DriftFile, fileMiddleware: FileMiddleware | undefined): RouteInfo[] {
+  const routes: RouteInfo[] = [];
+  const lines = file.content.split("\n");
+  const echoPattern = /\.\b(GET|POST|PUT|PATCH|DELETE|HEAD|OPTIONS)\s*\(\s*"([^"]+)"/;
+  const gorillaPattern = /HandleFunc\s*\(\s*"([^"]+)".*\.Methods\s*\(\s*"(\w+)"/;
+
+  for (let i = 0; i < lines.length; i++) {
+    const line = lines[i];
+    // Skip comment lines — prevents phantom routes from commented-out code.
+    if (isCommentLine(line, C_STYLE_COMMENT_MARKERS)) continue;
```
skhan75, line 34: Not a bug: this is identical to the original. Just flagging that it's completely untested.

Delete this line, or the matching one in `js.ts:27`, and the full suite still passes (1871/1871). Delete the Python equivalent in `python.ts:85` and a test fails.

So the `//` and `/* */` half of the #66 fix has no coverage at all, and it just moved into new files. This is the main reason I'm asking for the boundary test.

src/drift/route-extractors/shared.ts
```diff
@@ -0,0 +1,51 @@
...
+  return markers.some((m) => trimmed.startsWith(m));
+}
+
+// ─── Phase 2: inheritance resolution ─────────────────────────────────
+// A route's effective protection is its per-route middleware UNION the
+// file-level middleware detected for its file.
+export function inheritedAuth(perRoute: boolean, fileMw: FileMiddleware | undefined): boolean {
+  return perRoute || (fileMw?.hasAuth ?? false);
+}
+export function inheritedValidation(perRoute: boolean, fileMw: FileMiddleware | undefined): boolean {
```
skhan75, line 38: `inheritedValidation` here and `inheritedRateLimit` on line 41 are both untested. Change either one to just `return perRoute` (dropping the file-level part) and the suite stays green. `inheritedAuth` on line 35 does have a test.

These two feed the Input-validation and Rate-limiting findings users actually see, so worth pinning in the same test file.

skhan75 (commented):
Really nice extraction, the module split is the right one.

I checked this properly rather than just reading the diff: I built `main` and your branch side by side and compared their actual output. 51 routes pulled through the regex fallback across all four languages, byte for byte identical. Same for scans of three real repos. So the move itself is clean. ✅

One thing I'd like before merge, plus three small ones (left inline).

## The ask: the tests can't see most of what you moved

Your argument is "the existing suite is the guardrail." I tested that claim by deliberately breaking the moved code and seeing if the suite noticed. Mostly it didn't:

| What I broke | Suite result |
|---|---|
| deleted the comment-skip in `js.ts:27` | still green |
| deleted the comment-skip in `go.ts:28` | still green |
| deleted the comment-skip in `python.ts:85` | caught it ✅ |
| `perVal = false` in js / go / python | still green |
| `perRate = false` in js / go / python | still green |
| `inheritedValidation` / `inheritedRateLimit` gutted | still green |

11 out of 11 breakages went unnoticed.

Here's what that means concretely. I can delete this line from `js.ts` right now:

```ts
if (isCommentLine(lines[i], C_STYLE_COMMENT_MARKERS)) continue;
```

and every single test still passes. That's the fix from #66, the PR immediately before this one. Without it, this:

```js
// router.post('/admin/delete', h)   <- commented out
router.post('/admin/delete', h)      <- the real one
```

goes back to producing a phantom route, and nothing would tell us.

**None of this is your fault.** The gap already existed, and your code is byte-identical to the original. But you've just moved that code into four brand new files, and "the suite protects us" is the stated reason for shipping no tests. A copy/paste slip during the move would have been invisible.

**So: could you add `test/unit/drift/route-extractors.test.ts`?** Call the extractors directly on tree-less files and assert the whole object:

```ts
expect(goRouteExtractor.extract(file, deps)).toEqual([
  { method: "POST", path: "/users", file: "a.go", line: 4,
    hasAuth: true, hasValidation: false, hasRateLimit: false, hasErrorHandler: false },
]);
```

The key bit is `toEqual` on the full `RouteInfo`, not just `hasAuth`. A dozen cases like that would catch nearly everything in the table above, and it's exactly the artifact a "behavior-preserving refactor" should leave behind so the next person editing these files is safe.

## On the test failures you asked about

Confirmed, and they're fine. CI here is green on all 4 jobs, and the suite is green for me on macOS too. `concise-ai-summary` only fails under `FORCE_COLOR=1`, and `git-metadata` is git timing on Windows. Neither file is touched by this diff, so nothing to worry about.

## One thing I want to be clear about

Leaving `buildFileMiddlewareIndex` behind was the **right** call, not an unfinished job. #69 scoped it out, and untangling those cross-language guards would have been risky for zero behavioral gain. Good judgement.

Happy to approve once the test file lands.

---

🤖 Reviewed with [Claude Code](https://claude.com/claude-code)

src/drift/route-extractors/go.ts
BrandonGarate177: Done

src/drift/route-extractors/js.ts
BrandonGarate177: Also done

src/drift/route-extractors/python.ts
BrandonGarate177: Also also done

src/drift/security-consistency.ts
BrandonGarate177: Okay, it's now `Record<SupportedLanguage, RouteExtractor>`, so a future 6th language without an extractor is a compile error at the literal. 

Kept a runtime `in` guard so files whose `language` isn't a `SupportedLanguage` (or is null) are still skipped safely.

src/drift/route-extractors/types.ts
BrandonGarate177: yea you right, I deleted it

src/drift/route-extractors/go.ts
BrandonGarate177: The "skips a commented-out route" cases in `route-extractors.test.ts` assert exactly one route.

src/drift/route-extractors/shared.ts
BrandonGarate177: The JS "inherits validation + rate-limit from file-level middleware" case feeds a route with no inline signal and a `fileMw` entry, so both helpers are used directly

BrandonGarate177:
Lmk how this looks 😄  !!

skhan75 (approved):
Noice

PR #72, opened 2026-07-23
Feat/import style multilang clean

### LINKED ISSUE #56: Import-style / convention drift is JS/TS only; add Python, Go, Rust

BrandonGarate177:
## Summary

Extends the `import-consistency` detector beyond JS/TS to **Go, Python, and Rust**, and generalizes it into a language-agnostic, per-axis core with one stateless classifier per language behind a shared interface — mirroring the `route-extractors/` pattern from #70.

Previously `import-consistency` only understood JS/TS relative-vs-alias paths. It now detects, per language:

| Language | Axes |
|---|---|
| JS/TS | `path_style` (relative `./` vs alias `@/`) — ES `import … from` **and** CommonJS `require()` |
| Go | `go_grouping` (stdlib/external blank-line groups vs flat), `go_ordering` (gofmt byte-order within groups) |
| Python | `py_path_style` (absolute `from pkg.mod` vs relative `from .mod`), `py_wildcard` (`import *` vs explicit) |
| Rust | `rust_glob` (`use …::*` vs explicit), `rust_use_path` (`crate::` vs `super::`/`self::`), `rust_grouping` (origin-grouped vs flat) |

Each **axis** is voted independently, so a file can be consistent on one dimension and drift on another. Detector id (`import-consistency`), category (`import_style`), and scoring wiring are unchanged.

## Type of change

- [x] New feature
- [x] Refactor (generalized the detector; JS/TS behavior preserved)
- [x] Bug fix (intent-seed axis leak — see below)
- [x] Test

## Architecture

New `src/drift/import-style/` module, mirroring `route-extractors/`:

- `types.ts` — `ImportStyleClassifier` interface + `AxisClassification`.
- `labels.ts` — `AXES` metadata (single source of truth for axis copy + valid pattern keys), `Axis` / `PatternOf` types.
- `patterns.ts` — named, individually-testable regexes.
- `shared.ts` — `cleanTree`, `capEvidence`, `EVIDENCE_LIMIT`, `binaryMajority`.
- `js.ts` / `go.ts` / `python.ts` / `rust.ts` — one classifier each, AST-first with a regex fallback.

`import-consistency.ts` is now a thin dispatch over a total `Record<SupportedLanguage, ImportStyleClassifier>`.

## Technical decisions & implementation

- **Per-axis voting.** Each classifier returns `AxisClassification[]`; the core buckets classifications by axis and runs the existing directory-scoped dominance vote (min 3 files, 0.7 dominance, temporal weighting) plus the project-wide entropy gate **once per axis**. Axis keys are namespaced per language (`go_grouping`, `py_wildcard`, …) so votes never mix across languages, and the `import_style` scoring category is reused unchanged.
- **Dispatch + compile-time safety.** `CLASSIFIERS` is a *total* `Record<SupportedLanguage, ImportStyleClassifier>`, so adding a language to the union without a classifier is a compile error. `AxisClassification` is a distributive union where `pattern` is derived from each axis's `patternNames` via `PatternOf<A>` — pairing an axis with a foreign pattern is a compile error rather than an `undefined` label at runtime.
- **AST-first, regex fallback.** Classifiers use the tree-sitter parse when it's clean and fall back to a line/regex scan otherwise, so a syntactically-broken file still produces a best-effort signal instead of nothing.
- **JS/TS behavior preserved, plus CommonJS.** The ES-module path is byte-for-byte unchanged (verified by dogfooding — see Testing). CommonJS `require("./x")` specifiers are now counted too, so `require`-based repos (e.g. Express) are analyzed rather than silently skipped; comment lines are excluded to avoid counting commented-out imports.
- **Go reads every import form.** Single-line `import "x"`, block `import ( … )`, and multiple blocks are all collected (all `import_spec` nodes on the AST path; a stateful scan on the regex path). Grouping/ordering treat each blank-line group independently, matching how gofmt sorts within groups.
- **Rust glob idioms are excluded.** `use super::*;` (test re-imports) and `use self::Enum::*;` (enum-variant scoping) are near-universal idioms, and external/std `::prelude::*` globs are designed to be glob-imported — all excluded so they don't read as drift. `use crate::*;` / `use crate::prelude::*;` stay flagged as deliberate broad globs. This distinction came directly out of testing on real Rust repos, which were otherwise noisy.
- **Intent seed is per-axis.** An `import_style` intent hint (`alias`/`relative`) only describes `path_style`, so it is seeded **only** into the axis whose pattern vocabulary matches. Previously a single hint leaked into every axis, injecting a phantom pattern and bypassing the dominance threshold — emitting spurious divergences on unrelated axes. Fixed and covered by a regression test.

## Deferred (documented, not in scope here)

- **Python import ordering** — isort's category-first section ordering can't be validated shallowly without a stdlib list + first-party package inference, which would false-positive on correctly-sorted code. Deferred to a config/category-aware follow-up.

## Testing

- **Unit + calibration:** per-classifier tests (AST + regex fallback) for every axis; calibration fixtures confirm each primary axis fires on a mixed-convention directory and stays silent on a no-dominant one. Regression tests for the intent-seed leak (proven to fail without the fix) and the Rust idiom exclusions.
- **Full suite:** `npm run lint` / `typecheck` / `build` clean; **1948 passing**, with only 2 pre-existing environment-flaky failures unrelated to this change (`git-metadata` timing, `concise-ai-summary` ANSI).
- **Dogfood (behavior-preservation):** ran VibeDrift on itself on `main` vs this branch — the pre-existing JS/TS `path_style` findings are **byte-identical**; the only new findings are this branch's own added files.
- **Real-repo testing across all 5 languages:**
  - Correctness set (express, zod, gin, mux, requests, flask, ripgrep, bytes): no crashes; correctly silent on consistent codebases; Rust axes fire on genuinely-mixed code. This surfaced (and drove) the Rust idiom exclusions and CommonJS support.
  - Intense set (~35k files): three.js, react, TypeScript, angular, go, terraform, django, pandas, deno, vector, tikv. Zero crashes/timeouts. django produced legit `py_path_style`/`py_wildcard` findings; deno/vector/tikv produced directory-scoped Rust findings with the idiom exclusions holding at scale; consistent JS/TS/Go stayed silent.

## Checklist

- [ ] `README.md` updated — n/a (no flag/command/user-facing surface changed; detector id and category unchanged). Handbook + CHANGELOG updated instead.
- [x] This change improves how accurately VibeDrift measures drift
- [x] Copy uses "Vibe Drift Score"

## Related issues

Closes #56


skhan75:
This is a big change, allow me some time to review this @BrandonGarate177

BrandonGarate177:
Take your time !!

src/drift/import-style/go.ts
```diff
@@ -0,0 +1,119 @@
...
+  const t = text.trim();
+  if ((t.startsWith('"') && t.endsWith('"')) || (t.startsWith("`") && t.endsWith("`"))) return t.slice(1, -1);
+  return t;
+}
+
+/** Every import spec in the file, in source order — whether in a block
+ *  `import ( … )` (one or many blocks) or a single-line `import "x"`. The gap
+ *  between blocks/lines is a row-gap > 1, so grouping/ordering treat each as
+ *  its own group (matching gofmt). */
+function collectAst(tree: Tree): Spec[] {
```
skhan75, line 45: Is there an unit test that checks this specific functionality?

skhan75 (commented):
Really solid work on this one. The per-language classifier split mirrors route-extractors nicely, the regexes living in one testable file is exactly the lesson from #70, and the ES module path checks out as behavior preserving against main. We put this through a deep review pass (every finding below was independently verified with a concrete repro before writing it up), and there is one real bug to fix plus a handful of coverage gaps and small cleanups. None of it undermines the approach, the structure is right.

The one I would fix first is the intent-seed leak in import-consistency.ts, it can flag perfectly consistent Rust or Python code off the back of a JS-flavored CLAUDE.md line. Details inline.

🤖 Reviewed with [Claude Code](https://claude.com/claude-code)

src/drift/import-consistency.ts
```diff
@@ -1,139 +1,121 @@
...
+    // blanket seed across unrelated axes (which would inject a phantom pattern
+    // and bypass the dominance threshold on, say, go_grouping).
+    const hintPattern = pickIntentHint(ctx, "import_style")?.pattern;
+
+    for (const [axis, profiles] of byAxis) {
+      if (profiles.length < 3) continue;
+      const meta: AxisMeta = AXES[axis]; // total by construction — axis is a key of AXES
+
+      // Seed only the axis this declaration actually applies to.
+      const seededPattern = hintPattern && hintPattern in meta.patternNames ? hintPattern : undefined;
```
skhan75: This guard lets a JS instruction accidentally police Rust and Python code. Walk through it: someone's CLAUDE.md says "use relative imports", which is a JavaScript convention. But `relative` is also a pattern key on the Python axis and the Rust axis, so this membership check seeds all three. And a seeded axis skips the dominance threshold.

Concretely, take a Rust directory where every file agrees:

```rust
use crate::config::Config;   // all 4 files import exactly this way
```

With that CLAUDE.md line present, this code emits a `rust_use_path` finding with zero deviators on perfectly consistent Rust, purely because of an instruction about JavaScript. Remove the hint and nothing fires. The user sees a drift warning they cannot act on.

The regression test covers `alias`, which happens to be the one key that exists on only one axis, so it can't catch this. Mapping hints to axes explicitly (`import_style` hints bind only to `path_style`) instead of checking key membership fixes it, plus a `relative` version of the leak test.

src/drift/import-style/patterns.ts
```diff
@@ -0,0 +1,47 @@
...
+export const PY_WILDCARD = /^\s*from\s+\S+\s+import\s+\*/;
+
+// ─── Rust (glob + use-path-style axes) ───
+
+/** A `use` declaration line (optionally `pub use`). */
+export const RUST_USE = /^\s*(?:pub\s+)?use\s+/;
+/** A glob `use` — `use …::*`. */
+export const RUST_USE_GLOB = /use\s[^;]*::\*/;
+/** The head segment of a `use` path → capture [1] (`crate` | `super` | `self` | a crate name). */
+export const RUST_USE_HEAD = /^\s*(?:pub\s+)?use\s+(\w+)/;
```
skhan75: These two patterns only accept `pub` followed by a space, so the most common re-export forms in real crates never match:

```rust
pub(crate) use crate::db::Pool;   // silently invisible to the detector
pub(super) use self::helpers::*;
```

Three things go wrong: those lines drop out of `rust_use_path` voting entirely, the regex fallback skips them, and worst, `pub(crate) use super::*;` gets flagged as the glob anti-pattern because the idiomatic `super`/`self` exemption reads the head capture, which is undefined here.

Widening both patterns to `(?:pub(?:\s*\([^)]*\))?\s+)?` fixes all three. Also worth knowing: nothing currently tests even plain `pub use`, deleting `(?:pub\s+)?` leaves the whole suite green, so a test per visibility form would pin this down.

src/drift/import-style/rust.ts
```diff
@@ -0,0 +1,126 @@
...
+ *  single-origin file is never a false deviator. Note: rustfmt does not enforce
+ *  import grouping by default, so this is a softer convention than gofmt's. */
+function grouping(rows: UseRow[]): AxisClassification | null {
+  if (rows.length < 2) return null;
+  const origins = new Set(rows.map((r) => useOrigin(r.text)).filter((o): o is Origin => o !== null));
+  if (origins.size < 2) return null;
+  const sorted = [...rows].sort((a, b) => a.line - b.line);
+  let grouped = false;
+  for (let i = 1; i < sorted.length; i++) {
+    if (sorted[i].line - sorted[i - 1].line > 1) { grouped = true; break; }
```
skhan75: rustfmt itself can fool this check. It decides "grouped" from line-number gaps between where each `use` starts, but rustfmt wraps long import lists across lines by default:

```rust
use std::collections::HashMap;
use serde::{
    Deserialize,
    Serialize,
};
use crate::config::Config;
```

No blank line anywhere, this file is flat. But the third `use` starts 4 lines after the second one, so the gap rule says `grouped`. Every file with one wrapped import gets mislabeled, and in a flat-convention repo those become false deviators in the vote.

Since the AST path has `endPosition`, measuring the gap from the previous declaration's end row (and checking the line in between is actually blank) fixes it. Comments and `#[cfg]` attributes between uses trip the same wire.

src/drift/import-style/go.ts
```diff
@@ -0,0 +1,119 @@
...
+function evidenceOf(specs: Spec[]): { line: number; code: string }[] {
+  return capEvidence(specs.map((s) => ({ line: s.row + 1, code: s.code })));
+}
+
+function grouping(specs: Spec[]): AxisClassification | null {
+  if (specs.length < 2) return null;
+  if (new Set(specs.map((s) => s.category)).size < 2) return null;
+  let grouped = false;
+  for (let i = 1; i < specs.length; i++) {
+    if (specs[i].row - specs[i - 1].row > 1) { grouped = true; break; }
```
skhan75: The gap rule here can be fooled by a comment. This line decides "grouped" by checking whether the line numbers of two imports jump by more than 1, it never checks that the line in between is actually blank. A comment creates the same jump.

Concretely, this file is flat, no blank line anywhere:

```go
import (
    "fmt"
    // helpers
    "github.com/pkg/errors"
)
```

But `"fmt"` and `"github.com/pkg/errors"` are 2 line numbers apart, so this line classifies it as `grouped` (blank-line separated). In a repo where every file is flat, adding a harmless comment inside your imports makes that file a deviator in the vote, which is a false alarm from the user's point of view. gofmt attaches comments to the spec that follows them, only blank lines delimit groups.

The fix is to check the lines between the two specs are actually blank instead of just comparing row numbers, that covers both the AST and regex paths.

src/drift/import-style/js.ts
```diff
@@ -0,0 +1,56 @@
...
+    for (let i = 0; i < lines.length; i++) {
+      const line = lines[i].trim();
+      if (isCommentLine(line, C_STYLE_COMMENT_MARKERS)) continue;
+      if (JS_IMPORT_LINE.test(line)) {
+        // ES module: import … from "spec"
+        const fromMatch = line.match(JS_FROM_SPECIFIER);
+        if (fromMatch) record(fromMatch[1], line, i + 1);
+      } else {
+        // CommonJS: any number of require("spec") on the line.
+        for (const m of line.matchAll(JS_REQUIRE)) record(m[1], line, i + 1);
```
skhan75: The CommonJS branch scans every line of the file with `matchAll(JS_REQUIRE)`, and a `require(...)` inside a comment or string counts too:

```js
const a = loadA("x"); // was require("./legacy-a")
const b = loadB("y"); // was require("./legacy-b")
const c = loadC("z"); // was require("./legacy-c")
```

Three imports that don't exist, but that's enough for the file to classify as `relative` (isCommentLine only skips lines that START with a comment marker, trailing comments pass through). The old detector never looked at non-import lines, so this false-positive surface is new in this PR, not inherited.

Stripping a trailing `//` comment from the line before the matchAll kills the common case cheaply. And credit where due, the ES module path checks out as behavior preserving against main.

src/drift/import-style/js.ts
```diff
@@ -0,0 +1,56 @@
...
+
+    const lines = file.content.split("\n");
+    let relativeCount = 0;
+    let aliasCount = 0;
+    const evidence: Evidence[] = [];
+
+    // Count a single specifier: relative (./ ../) vs alias (@/ ~/); external /
+    // bare packages have no local path-style choice and are skipped.
+    const record = (spec: string, code: string, lineNo: number) => {
+      if (spec.startsWith("./") || spec.startsWith("../")) relativeCount++;
```
skhan75, line 29: The header says this matches the original analyzeImports, but three of its arms have no test holding them in place. I tried each of these mutations and the full suite stayed green every time:

1. Drop `(?!type\s)`, so `import type { X } from "./x"` starts counting toward the threshold (it must not).
2. Delete the `../` arm, so parent-relative imports are ignored completely.
3. Delete the `~/` arm, so that alias style disappears.

The existing tests only ever use `./` and `@/` specifiers, which is why all three survive. Three one-line test cases, one per arm, would turn the parity claim in this header into something the suite actually proves.

src/drift/import-style/python.ts
```diff
@@ -0,0 +1,141 @@
...
+
+function pathStyleFromRegex(lines: string[], pkg: Set<string>): AxisClassification | null {
+  const c: PathCounts = { relative: 0, absoluteLocal: 0, evidence: [] };
+  for (let i = 0; i < lines.length; i++) {
+    const line = lines[i];
+    let kind: "relative" | "absoluteLocal" | null = null;
+    if (PY_FROM_RELATIVE.test(line)) kind = "relative";
+    else {
+      const m = line.match(PY_FROM_ABSOLUTE);
+      if (m && pkg.has(m[1].split(".")[0])) kind = "absoluteLocal";
```
skhan75, line 76: The `pkg.has(...)` gate here is what stops third-party imports from voting, but nothing tests it. Concretely, picture a Python file with a syntax error (so no AST, regex fallback takes over) containing only:

```python
from os import path
from django.db import models
```

Neither is a project import, so this file should stay silent. Replace the gate with plain `if (m)` and the entire suite still passes, meaning that protection is currently unpinned: if it regressed, files like this would classify as "absolute" and could show up as deviators in someone's report.

The AST path has exactly this negative test already, a treeless mirror of it would cover this arm in a few lines.

src/drift/import-style/python.ts
```diff
@@ -0,0 +1,141 @@
...
+import { EVIDENCE_LIMIT, capEvidence, cleanTree, binaryMajority } from "./shared.js";
+import { isCommentLine, PYTHON_COMMENT_MARKERS } from "../comment-markers.js";
+
+/**
+ * Blank out `#` comment lines and triple-quoted docstring bodies so the
+ * from-import regexes never match an import that only appears inside a comment
+ * or a docstring. Regex-fallback only — the AST path is already immune
+ * (a docstring is a string literal, not an `import_from_statement`).
+ */
+function stripPyNonCode(lines: string[]): string[] {
```
skhan75, line 33: This re-implements the triple-quote docstring tracker that `route-extractors/python.ts` already carries, and the two copies have already drifted apart on the simplest case:

```python
"""one-line docstring"""
```

This copy blanks that line (the `triples >= 2` branch), the route-extractor copy falls through and can still match inside it. Two modules, same job, different answers.

Since this PR introduced `comment-markers.ts` precisely so these modules share their "is this line code?" logic, the docstring state machine belongs there too, used by both callers, before the copies drift further.

src/drift/import-style/patterns.ts
```diff
@@ -0,0 +1,47 @@
...
+
+/** Start of a block import: `import (`. */
+export const GO_IMPORT_BLOCK_START = /^\s*import\s*\(/;
+/** End of a block import: a line whose first non-space char is `)`. */
+export const GO_IMPORT_BLOCK_END = /^\s*\)/;
+/** A quoted import path inside a spec line → capture [1] (double) or [2] (raw backtick). */
+export const GO_IMPORT_PATH = /"([^"]+)"|`([^`]+)`/;
+/** A single-line import: `import "fmt"`, `import _ "x"`, `import . "x"`,
+ *  `import alias "x"` → capture [1] = path. Never matches `import (`. */
+export const GO_IMPORT_SINGLE = /^\s*import\s+(?:(?:[A-Za-z_]\w*|\.|_)\s+)?["`]([^"`]+)["`]/;
```
skhan75, line 27: The alias arm of this pattern is documented but nothing exercises it. These are all real single-line Go imports:

```go
import _ "github.com/lib/pq"     // blank-identifier driver import
import . "testing"
import myalias "github.com/a/b"
```

Removing `(?:(?:[A-Za-z_]\w*|\.|_)\s+)?` from the pattern passes the full suite, so the regex fallback silently losing all three forms would ship unnoticed. The `import _ "..."` driver form is the most common single-line import in real Go code, so one treeless test with it would pin this arm.

src/drift/import-style/labels.ts
```diff
@@ -0,0 +1,99 @@
+/**
+ * Per-axis copy + pattern labels. Each `AxisClassification.axis` a classifier
+ * emits must have an entry here; the detector uses it to build finding text.
+ */
+
+export interface AxisMeta {
+  /** subCategory on the finding. */
+  subCategory: string;
```
skhan75: `subCategory` is set to exactly its own axis key in all eight entries, and both readers already have `axis` in scope, so the field is dead weight, the same shape as the dead `language` field we caught on route-extractors in #70. The risk is that it's a plain string restated eight times:

```ts
{ axis: "py_wildcard", subCategory: "py_wilcard", ... }  // typo ships silently
```

Using the axis key directly at the two read sites makes that mistake impossible. Smaller nit while you're here: `headline` and `axisLabel` are identical in seven of eight entries, one field would probably do.

test/calibration/import-style.test.ts
```diff
@@ -0,0 +1,59 @@
...
+    totalLines: driftFiles.reduce((s, f) => s + f.lineCount, 0),
+    dominantLanguage: lang,
+  };
+}
+
+function forAxis(ctx: DriftContext, sub: string) {
+  return importConsistency.detect(ctx).filter((f) => f.subCategory === sub);
+}
+
+const CASES = [
```
skhan75, line 39: The header says this gate proves the two behaviors "for each new-language axis", but CASES covers three of the seven: go_grouping, py_path_style, rust_glob. The other four (go_ordering, py_wildcard, rust_use_path, rust_grouping) never pass through `importConsistency.detect` in any test, so their vote wiring, entropy gating and finding copy are unexercised end to end.

Your fixture pattern makes this cheap to close, it's one file factory per axis. Either add the four, or reword the header to say it samples one axis per language, so the comment doesn't claim more than the suite proves.

CHANGELOG.md
```diff
@@ -37,6 +37,17 @@ explicitly under **Breaking** so CI users can recalibrate.
 
 ## 0.16.3 — 2026-07-22
 
+### Added
+
+- **Import-style drift now covers Go, Python, and Rust** (previously JS/TS only). The
```
skhan75: Small bookkeeping one: this entry sits inside the `## 0.16.3` section, and 0.16.3 is already published on npm without this feature (the 0.17.0 notes are above it, and main's package.json is already 0.17.0). Anyone reading the changelog would believe a version they already have includes multi-language import-style.

Could you move the bullet to a new unreleased/next-version section at the top? The entry text itself is fine as is.

src/drift/import-style/go.ts
```diff
@@ -0,0 +1,119 @@
...
+  const specs: Spec[] = [];
+  let inBlock = false;
+  for (let i = 0; i < lines.length; i++) {
+    if (isCommentLine(lines[i], C_STYLE_COMMENT_MARKERS)) continue; // a commented "path" is not a real import
+    if (inBlock) {
+      if (GO_IMPORT_BLOCK_END.test(lines[i])) { inBlock = false; continue; }
+      const m = lines[i].match(GO_IMPORT_PATH);
+      if (m) specs.push({ row: i, path: m[1] ?? m[2], category: goCategory(m[1] ?? m[2]), code: lines[i].trim() });
+    } else if (GO_IMPORT_BLOCK_START.test(lines[i])) {
+      inBlock = true;
```
skhan75, line 68: One more in the fallback path, easiest to see with a file:

```go
import ( "fmt" )

func main() { fmt.Println("hello world") }
```

The block-start regex matches line 1 and sets `inBlock`, but the path on that same line is never captured, and since the `)` is also on that line, the block-end check never fires. So the scanner stays "inside the import block" for the whole file and collects every quoted string it meets. I reproduced a `go_grouping` classification whose evidence line was literally `fmt.Println("hello world")`.

Fallback-only, but the output is garbage when it happens. Handling a path and a closing `)` on the block-start line closes it.

src/drift/import-consistency.ts
BrandonGarate177: Ok this makes sense, i fixed it by binding the `import_style` hint straight to `path_style` instead of checking key membership. 

I also added `relative`-variant leak tests for python and rust. the og `alias` test only lives in one axis.

src/drift/import-style/go.ts
BrandonGarate177: It's in `import-style-go.test.ts`. I made sure we actually guard it - > reverting `collectAst` to first block only turns 4 of them red.

src/drift/import-style/patterns.ts
BrandonGarate177: I widened both to `(?:pub(?:\s*\([^)]*\))?\s+)?` so that these : 

- `pub(crate)` 
- `pub(super)` 
- `pub(in path)` 

would all match. 

Also added a test per visibility form.

src/drift/import-style/rust.ts
BrandonGarate177: Oh I actually didn't think about this, the row-gap was lying to me. I started by tracking each use's *end* row and only calling it grouped when there is a blank line between declerations. `blankBetween`.

src/drift/import-style/go.ts
BrandonGarate177: Same idea as the rust one from above. I was comparing row numbers instead of just checking if the line below was empty.

src/drift/import-style/js.ts
BrandonGarate177: I fixed by stripping a trailing `//` comment before the `matchAll`

src/drift/import-style/js.ts
BrandonGarate177: I added one test each: `import type … from "./x"` doesn't count toward the threshold, `../` counts as relative, and `~/` counts as alias.

src/drift/import-style/python.ts
BrandonGarate177: Ok, so i'm noticing that i forget writting tests a lot here lol

BrandonGarate177: I added a treeless test with just `from os import path` / `from django.db import models` and asserted it stays silent , swap the gate for `if (m)` and it goes red.

src/drift/import-style/python.ts
BrandonGarate177: I moved the triple-quote docstring machine into `comment-markers.ts` as `pythonNonCodeLines`, and both python fallbacks (import-style + route-extractors) use it now

src/drift/import-style/patterns.ts
BrandonGarate177: I added a treeless `import _ "github.com/lib/pq"` case (the blank-identifier driver form) to pin it.

src/drift/import-style/labels.ts
BrandonGarate177: I dropped it and the detector uses the axis key directly now, so a typo can't ship. I also merged `headline`/`axisLabel` into one `label`

test/calibration/import-style.test.ts
BrandonGarate177: I went with adding coverage rather than softening it: fixtures + cases for `go_ordering`, `py_wildcard`, `rust_use_path`, and `rust_grouping`, so all seven axes run through the detector now.

CHANGELOG.md
BrandonGarate177: I moved the bullet up into a fresh `## [Unreleased]` section; 0.16.3 is untouched.

src/drift/import-style/go.ts
BrandonGarate177: Fixed it by having the block-start branch grab a path on that line and close the block when the `)` is on the same line, so it stops leaving `inBlock` open and collecting every quoted string after it

BrandonGarate177:
Ok, let me know how you think this looks please 😄  !

src/drift/import-style/go.ts
```diff
@@ -0,0 +1,137 @@
...
+      // A one-line block like `import ( "fmt" )`: capture the path and, if the
+      // closing `)` is on the same line, close the block immediately — otherwise
+      // inBlock stays open and every later quoted string is slurped as a spec.
+      const rest = lines[i].replace(GO_IMPORT_BLOCK_START, "");
```
skhan75, line 72: This is good

src/drift/import-style/python.ts
skhan75: Fix looks right. One last thing to check is the one-line docstring case `(\"\"\"like this\"\"\")` is the spot where the two old copies disagreed, and nothing pins it yet. One tiny test on pythonNonCodeLines and I'm happy to resolve.

src/drift/import-style/python.ts
skhan75: way to go buddy

skhan75:
Ran a deeper Rust-focused pass on `src/drift/import-style/rust.ts` (this classifier will judge a lot of real-world crates, so it earned one). Seven inline comments just posted, each with a small runnable example. The two on the collector (test modules poisoning the vote) are the ones I would fix before merge, the rest can be fast-follows. Every finding was reproduced against the classifier before writing it up.

🤖 Reviewed with [Claude Code](https://claude.com/claude-code)

src/drift/import-style/rust.ts
```diff
@@ -0,0 +1,138 @@
...
+
+interface UseRow { start: number; end: number; text: string; } // 0-based start/end rows
+
+/** Collect one row per `use` declaration: start/end line (multiline uses span
+ *  several rows) plus the single-line text of the head. */
+function collectUses(file: DriftFile): UseRow[] {
+  const rows: UseRow[] = [];
+  const tree = cleanTree(file);
+  if (tree) {
+    for (const u of tree.rootNode.descendantsOfType("use_declaration")) {
```
skhan75: This collector picks up `use` statements from the whole file at any depth, and that makes the standard test idiom poison the vote. Nearly every Rust file ends with:

```rust
use std::fmt;
use std::io;

#[cfg(test)]
mod tests {
    use super::*;
}
```

That inner `use super::*;` counts as a second origin, and the blank line before `mod tests` reads as a group separator, so this flat file classifies as `grouped`. I reproduced it: evidence lines [1, 2, 8]. In a flat-convention repo, virtually every file with tests becomes a false deviator.

Counting only top-level uses (parent is `source_file`), or only the contiguous import block, fixes it for the grouping axis.

src/drift/import-style/rust.ts
```diff
@@ -0,0 +1,138 @@
...
+function glob(rows: UseRow[]): AxisClassification | null {
+  // Idiomatic globs are neither the glob anti-pattern nor "explicit" — drop them.
+  const relevant = rows.filter((r) => !isIdiomaticGlob(r.text));
+  const globRows = relevant.filter((r) => RUST_USE_GLOB.test(r.text));
+  if (globRows.length === 0 && relevant.length < 2) return null;
+  const evidence = capEvidence((globRows.length > 0 ? globRows : relevant).map((r) => ({ line: r.start + 1, code: r.text })));
+  return { axis: "rust_glob", pattern: globRows.length > 0 ? "glob" : "explicit", evidence };
+}
+
+function usePath(rows: UseRow[]): AxisClassification | null {
```
skhan75, line 130: Same root cause, different axis: test modules conventionally import helpers via `crate::` no matter what style the file itself uses, and those votes count here.

```rust
use super::config::Config;
use super::db::Db;

#[cfg(test)]
mod tests {
    use crate::test_utils::a;
    use crate::test_utils::b;
    use crate::test_utils::c;
}
```

The file's real style is relative (2 `super::` uses), but the 3 test-module `crate::` uses win the majority and it classifies as `crate`. Reproduced, with the `super::` lines showing up in the evidence of the wrong label. Restricting or down-weighting nested uses in collectUses fixes both this and the grouping comment above.

src/drift/import-style/rust.ts
```diff
@@ -0,0 +1,138 @@
...
+
+/** Collect one row per `use` declaration: start/end line (multiline uses span
+ *  several rows) plus the single-line text of the head. */
+function collectUses(file: DriftFile): UseRow[] {
+  const rows: UseRow[] = [];
+  const tree = cleanTree(file);
+  if (tree) {
+    for (const u of tree.rootNode.descendantsOfType("use_declaration")) {
+      if (!u) continue;
+      rows.push({ start: u.startPosition.row, end: u.endPosition.row, text: u.text.split("\n")[0].trim() });
```
skhan75: Keeping only the first line of the node text makes glob detection depend on formatting rather than meaning. Two real files:

```rust
use tokio::{
    io::*,
    net::TcpStream,
};
```

```rust
use std::io::{self, *};
```

Both contain genuine globs, both classify `explicit`. The first because the `*` sits on a continuation line this split throws away (and rustfmt produces exactly this wrapping), the second because the brace-group form is `, *` rather than the `::*` the regex requires. The header comment even mentions the `use_wildcard` node; querying `u.descendantsOfType("use_wildcard")` on the AST path fixes both in one line.

src/drift/import-style/rust.ts
```diff
@@ -0,0 +1,138 @@
...
+    }
+  } else {
+    const lines = file.content.split("\n");
+    for (let i = 0; i < lines.length; i++) {
+      if (RUST_USE.test(lines[i])) rows.push({ start: i, end: i, text: lines[i].trim() });
```
skhan75: The fallback has no comment or string guard, so dead code votes. A parse-error file containing:

```rust
/*
use crate::old_api::Widget;
use crate::old_api::Gadget;
*/
```

emits `rust_use_path = "crate"` with evidence pointing at commented-out code (the AST path on the same content correctly returns nothing). Trailing comments count too: `use crate::x::Foo; // was: use crate::x::*` classifies the file as glob style, because the unanchored glob regex happily matches inside the comment. Raw strings and macro_rules bodies vote the same way.

Truncating each row at the first `;` kills the trailing-comment case, and a small block-comment tracker (comment-markers.ts has no Rust support yet) covers the rest.

src/drift/import-style/rust.ts
```diff
@@ -0,0 +1,138 @@
...
+    }
```
skhan75, line 87: One more fallback divergence: rows record end = start, so the blankBetween fix from the earlier review only landed on the AST path. This file:

```rust
use std::collections::{
    HashMap,

    HashSet,
};
use crate::foo::Bar;
```

is flat on the AST path (the blank line is inside the brace group), but the fallback scans from the head line and sees that blank as a group boundary, so the same file classifies `grouped` treeless. Extending the fallback row's end past continuation lines until the closing `;` when the head opens a brace group brings the two paths back in line.

src/drift/import-style/rust.ts
```diff
@@ -0,0 +1,138 @@
...
+  let crate = 0;
+  let relative = 0;
+  const evidence: Evidence[] = [];
+  for (const r of rows) {
+    if (isIdiomaticGlob(r.text)) continue; // idiomatic glob — not a considered path-style choice
+    const head = r.text.match(RUST_USE_HEAD)?.[1];
+    const kind = head === "crate" ? "crate" : (head === "super" || head === "self") ? "relative" : null;
+    if (!kind) continue; // external crate — neutral
+    if (kind === "crate") crate++; else relative++;
+    if (evidence.length < EVIDENCE_LIMIT) evidence.push({ line: r.start + 1, code: r.text });
```
skhan75: Small one: evidence is pushed for every counted row before the majority is known, so it can showcase the losing side.

```rust
use super::helpers::x;
use crate::models::User;
use crate::db::Session;
```

This classifies `crate`, but evidence[0] is line 1, the `super::` use, the opposite pattern. To a user reading the report, a "crate-style" finding whose first evidence line is relative style reads as a wrong answer. Collecting the two sides separately and attaching only the winner's lines after the majority resolves fixes it.

src/drift/import-style/patterns.ts
```diff
@@ -0,0 +1,49 @@
...
+// ─── Rust (glob + use-path-style axes) ───
+
+/** A `use` declaration line, optionally with any visibility — `pub use`,
+ *  `pub(crate) use`, `pub(super) use`, `pub(in path) use`. */
+export const RUST_USE = /^\s*(?:pub(?:\s*\([^)]*\))?\s+)?use\s+/;
+/** A glob `use` — `use …::*`. */
+export const RUST_USE_GLOB = /use\s[^;]*::\*/;
+/** The head segment of a `use` path → capture [1] (`crate` | `super` | `self` | a crate name).
+ *  Tolerates any visibility prefix (`pub`, `pub(crate)`, …). */
+export const RUST_USE_HEAD = /^\s*(?:pub(?:\s*\([^)]*\))?\s+)?use\s+(\w+)/;
```
skhan75: Tiny valid-Rust gap: `RUST_USE_HEAD` requires a word character right after `use`, so the leading-:: disambiguation form has no head and vanishes:

```rust
use ::std::mem;
use ::serde::Deserialize;
```

These drop out of `rust_use_path` and can make `rust_grouping` undecidable by collapsing origins. It's a silent miss rather than a false alarm, so lowest priority of the batch, and the fix is one token: `use\s+(?:::)?(\w+)`.

src/drift/import-style/rust.ts
skhan75: This one is critial

skhan75 (changes requested):
Solid work! Few critical fixes on the Rust import class, and we are good to go

src/drift/import-style/python.ts
BrandonGarate177: I added a `pythonNonCodeLines` test in `drift/comment-markers.test.ts` that pins the one line """...""" where the two old copies diverge. The file also covers a multi line docstring and a # comment.

src/drift/import-style/rust.ts
BrandonGarate177: I fixed it by collecting **top-level uses only** so a `use super::*;` inside `#[cfg(test)] mod tests { ...}` no longer adds a phantom origin or a fake blank line. I added the test to `drift/import-style-rust.test.ts`

src/drift/import-style/rust.ts
BrandonGarate177: I believe the top level correction actually fixes this one too, the file votes on its real top level style. Our test uses the `super::` file + a `crate::` test mod and asserts `relative`.

src/drift/import-style/rust.ts
BrandonGarate177: I switched the glob detection to use `use_wildcard` AST node instead of first-line regex, so a wrapped `use tokio::{ io::*, … }` and a brace-group `use std::io::{self, *}` are both caught now.

test `drift/import-style-rust.test.ts`

src/drift/import-style/rust.ts
BrandonGarate177: Yea the fallback votes on dead code here, I added a `/* */` block comment treacker and strip the the trailing `//` before glob detection, so sommented-out uses and a `// .. use crate::x::*` note don't count 

`tests/unit/drift/import-style-rust.tests.ts`

src/drift/import-style/rust.ts
BrandonGarate177: I now follow each use across continuation lines to its `;` and record that as `end`, so a blank inside a wrapped brace group reads `flat` treeless too, matching the AST path.

src/drift/import-style/rust.ts
BrandonGarate177: I collect the `crate` and `relative` evidence separately now and attach only the winning side's lines after the majority resolves.

src/drift/import-style/patterns.ts
BrandonGarate177: I widened `RUST_USE_HEAD` to `use\s+(?:::)?(\w+)` so `use ::std::mem;` resolves a head instead of dropping out

src/drift/import-style/rust.ts
```diff
@@ -0,0 +1,158 @@
...
+ *  axes ignore them:
+ *   - relative: `use super::*;` (test re-imports), `use self::…::*;` (enum scoping)
+ *   - external prelude: `use rayon::prelude::*;` (preludes are designed to be glob-imported)
+ *  Crate-root (`use crate::*;`) and crate-internal (`use crate::prelude::*;`)
+ *  globs are deliberate local choices and stay flagged. */
+function isIdiomaticGlob(row: UseRow): boolean {
+  if (!row.isGlob) return false;
+  const head = headOf(row.text);
+  if (head === "super" || head === "self") return true;
+  return head !== "crate" && /\bprelude\b/.test(row.text);
```
skhan75: Almost there. The glob fix works, but it broke its own exemption.

This line checks for "prelude" on `row.text`, and `row.text` is only the first line of the declaration. But the new `use_wildcard` check now finds globs on the continuation lines too. So this gets flagged:

```rust
use rayon::{
    prelude::*,
    iter::IntoParallelRefIterator,
};
```

The glob is detected, then the exemption looks at `use rayon::{`, sees no prelude, and reports it as the anti-pattern. Before this PR there was no finding here at all, so it's a new false positive. Ran it over real crates to be sure: 7 hits across 5670 files, winnow and find-msvc-tools both flip.

Fix is to test the exemption against the full declaration text instead of `row.text`. Everything else in this batch looks good, this is the last one from me.

🤖 Reviewed with [Claude Code](https://claude.com/claude-code)

skhan75: I think, this was skipped until the glob fix was made, which is how it was hiding. Looks like one last fix. Rest all looks good

BrandonGarate177:
I pushed a couple more commits, and ran them over 12 crates. 

- **`274f64f`** — your prelude catch. Tested the exemption on the full declaration instead of just the first line, so a wrapped `use rayon::{ prelude::*, … }` isn't flagged anymore. Verified on winnow's `s_expression/parser.rs` + `testsuite/multiline.rs`. (`describe("Rust review round 3 …")`)
- **`cc75716`** — noticed `pub use self::…` re-exports (all over `mod.rs`) were counting as "relative" use-path votes, making mod.rs the lone deviator. They're re-exports, not import choices, so I skip `pub`/`pub(crate) use`. tokio use_path 8→2, clap 3→0. (`describe("… precision fix A")`)
- **`7d29a17`** — `use std::…::Ordering::*` was flagged but `use self::Ordering::*` wasn't. Enum-variant scoping is idiomatic, so I exempt it regardless of head (CamelCase target ⇒ enum, off the `use_wildcard` node). Bare `{…, *}` still flags. (`describe("… precision fix B")`)
- **`aa55db5`** — grouping findings pointed at the file's first `use`, not the line that actually shows the grouping. Now they show the pair around the blank boundary (or the run-together pair for flat), Rust + Go. Evidence only. (`describe("… fix C")`)

## Audit (ripgrep, tokio, serde, clap, alacritty, fd, winnow, regex, rustls, hyperfine, tracing, bat)

| axis | before → after |
|---|---|
| glob | 12 → **9** |
| use_path | 3 → 3 |
| grouping | 24 → 24 |
| new findings | **0** |


## Left alone on purpose
- serde's `use crate::lib::*;` still flags — by design, we keep crate-internal module globs flagged.
- The prelude check is "glob + contains `prelude`", so a contrived `use foo::{ prelude::X, bar::* }` would over-exempt. Doesn't happen in real code.
- Enum heuristic keys on CamelCase; a CamelCase-named module would over-exempt, but clippy won't let that exist.

src/drift/import-style/rust.ts
BrandonGarate177: The exemption now tests the full declaration text (row.full) instead of row.text:

  `return head !== "crate" && /\bprelude\b/.test(row.full);`

skhan75 (approved):

PR #92, opened 2026-08-02
fix: stop reporting an ORM that isn't there

### LINKED ISSUE #87: [bug] Data-access classifiers label routers / HTTP clients as "ORM" via unanchored \`ent.\` and CRUD-verb method regexes
## What happened

Two independent paths can stamp a file (or a proposed body) with an ORM data-access label when no ORM exists:

1. **Code DNA pattern classifier** (`src/codedna/pattern-classifier.ts`) — feeds `codedna-pattern` findings.
2. **Architectural-contradiction / data-access axis** (`src/drift/architectural-contradiction.ts`) — feeds directory-scoped dominance votes and the MCP `validate_change` tool via `classifyDataAccessLabel` (rendered label: `"ORM methods"`).

Both are plain regex over raw file text. Neither resolves the receiver, checks imports, or requires a known ORM client. Two specific patterns make an autorouter / HTTP-routing file a near-certain false positive:

**A. Unanchored `ent.` alternative (pattern-classifier only).** The "ORM import/usage" regex is:

```ts
/(?:gorm|prisma|sequelize|typeorm|sqlalchemy|django\.db|ent\.)\.?/i
```

`ent\.` was meant to catch Facebook/Ent's Go ORM (`ent.Client`, `ent.User.Query()`), but with no word boundary it matches **any identifier ending in `ent` followed by a dot**: `client.`, `httpClient.`, `component.`, `current.`, `content.`, `agent.`, `environment.`, `argument.`, `management.`, etc. A router that talks to an HTTP `client` is enough.

**B. CRUD-verb method names collide with router DSLs (both paths).** The drift detector uses:

```ts
/\.(?:Where|Find|Create|Save|Delete|First|Preload|findOne|findMany|findAll|objects\.filter)\s*\(/g
```

Those are GORM/Prisma/Sequelize vocabulary, but they are also natural names for route registration and route lookup (`router.Create("/users", h)`, `routes.findOne(pattern)`, `router.Delete(...)`). Method name alone is enough — no `prisma`/`gorm`/`db` receiver required.

**C. Path gate treats `*router*` as a handler file.** `isHandlerOrServiceFile` matches the substring `route` unanchored, so `autorouter.ts`, `router.go`, and `AutoRouter.ts` all enter the Bayesian classifier even when they are not handlers.

Practical effect: a file with zero ORM imports can still get classified as `orm` / `"ORM methods"`, then participate in (or become the minority of) a data-access dominance vote — exactly the PH report of "labeled our autorouter as ORM, there's no ORM in the repo."

## Steps to reproduce

1. Create a minimal project with a dominant repository-pattern handler directory (3+ files using an injected repo) and one autorouter with **no** ORM imports, e.g. `src/autorouter.ts`:
   ```ts
   export function mount(router: Router, client: HttpClient) {
     router.Create("/users", listUsers);
     router.Delete("/users/:id", deleteUser);
     const match = routes.findOne(req.path);
     return client.Get(match.url);
   }
   ```
2. Run `vibedrift . --format json` (or call `classifyDataAccessLabel` / the pattern classifier on that file alone).
3. Observe the autorouter's data-access / pattern label.

Alternatively, exercise the regexes in isolation (no scan required):

```js
const ormImport = /(?:gorm|prisma|sequelize|typeorm|sqlalchemy|django\.db|ent\.)\.?/i;
ormImport.test("client.Get(url)");           // true — no Ent ORM
ormImport.test("httpClient.Do(req)");        // true
ormImport.test("agent.RoundTrip");           // true

const ormMethods = /\.(?:Where|Find|Create|Save|Delete|First|Preload|findOne|findMany|findAll|objects\.filter)\s*\(/g;
ormMethods.test('router.Create("/users", h)'); // true — no ORM
ormMethods.test("routes.findOne(pattern)");    // true

/(?:handler|controller|service|route|endpoint|api|resource)/i.test("src/autorouter.ts"); // true
```

## Expected behavior

A file with no ORM import and no ORM-client receiver should not be labeled `orm` / `"ORM methods"`. At minimum:

- `ent.` should be word-bounded / package-anchored (`\bent\.` or an import-table check), not a bare three-letter suffix.
- CRUD-verb hits should require an ORM-shaped receiver or a known ORM import, not any `.Create(`/`.findOne(`.
- `*router*` path names should not automatically enter the handler/service classifier via the `route` substring.

## Actual behavior

- `client.Get(url)`, `httpClient.Do(req)`, `component.Render()`, `current.Path`, `agent.RoundTrip`, `content.Write(w)` all match the ORM import/usage regex.
- `router.Create(...)`, `routes.findOne(...)`, `routes.findAll(...)`, `router.Delete(...)` all match the ORM method regex used by `architectural-contradiction.ts`.
- `src/autorouter.ts` / `src/router.go` / `src/AutoRouter.ts` all pass `isHandlerOrServiceFile`.
- `classifyDataAccessLabel` can return `"ORM methods"` for a router-only body; that label is the same string surfaced through `validate_change`.

Confirmed by direct regex exercise against current `main` (0.19.1). No fixture or golden in `test/fixtures/classifier-bodies.txt` / `test/unit/drift/classifier-golden.test.ts` covers a router-shaped body — the only ORM golden is a real Prisma example.

## Environment

- VibeDrift version: 0.19.1
- Node version: v22.22.0
- OS: macOS (darwin arm64)

## Additional context

Handbook Ch 05 / Ch 07 describe these axes as "regex evidence extraction" — so *some* imprecision is an acknowledged design tradeoff. This specific failure mode is not documented as a known limitation (unlike other FP gaps tracked in `todo.md`), and the `ent\.` alternative in particular reads as a missing word-boundary bug rather than an intentional recall tradeoff.

Suggested fixes (independently useful):

1. Change `ent\.` to `\bent\.` (or require an `ent` import / `ent.` package qualifier from a resolved import table).
2. Gate ORM method-name hits on a known ORM receiver or import (`prisma|gorm|sequelize|typeorm|sqlalchemy|ent|objects`).
3. Tighten `isHandlerOrServiceFile` so `route` matches path segments (`/routes/`, `route.`, `-route.`) rather than the substring inside `router` / `autorouter`.
4. Add a golden fixture: autorouter / HTTP-client body that must classify as `none` or `http_client`, never `orm`.

skhan75:
## Summary

Three regex defects made VibeDrift report an ORM in codebases that have none. Scanning this repository against itself said `src/auth/api.ts uses http_client while 5/6 files use orm` at confidence 1.0, with no ORM in `package.json`. The label was inverted: the one file that genuinely is an HTTP client was reported as the deviation from a majority that does not exist.

Three commits, each independently measured and reviewable on its own:

**1. `fix(codedna): anchor the Ent ORM signal`** — the ORM regex carried a bare `ent\.` alternative with no left boundary, so any identifier ending in the letters "ent" plus a dot was read as Ent ORM usage. `file.content.split(...)` and `fullContent.indexOf(...)` were enough. That one alternative produced **125 of 125** "ORM import/usage" signals across five repos that declare no ORM at all.

A left `\b` alone is not sufficient: it still fires on a local variable named `ent` (`ent.ty`, `ent.entitled`). Real Ent usage is a package selector followed by an exported Go identifier, so the Ent alternative is split out as `/\bent\.[A-Z]/`, case-sensitive because the `/i` flag would make the `[A-Z]` discriminator a no-op.

**2. `fix(codedna): match handler paths by segment, not substring`** — `isHandlerOrServiceFile` matched unanchored substrings, so "route" inside `autorouter.ts`, `itty-router.js` and this repo's own `src/drift/route-extractors/` admitted files that are not handlers, and "api" admitted `therapist.ts` and `capital.ts`. A file that enters the classifier gets labelled by whatever signal fires first, so a parser was being given a data-access architecture label.

**3. `fix(drift): stop reading route registrations as ORM calls`** — the data-access detector matched bare CRUD verb names with no receiver and no import corroboration, so `router.Create("/users", h)` and `r.Delete("/orders/{id}", h.Del)` were evidence of an ORM. 7 of 8 realistic router bodies were labelled "ORM methods". This one is not confined to a scan report: `classifyDataAccessLabel` is the MCP `validate_change` contract and `hasDataAccessPattern` is on the session resolve path, so an agent editing a router mid-session could be told its change conflicts with an ORM the repo does not have.

## Before / after

before and after

A 7-file TypeScript service with no ORM in `package.json`, where five files contain `const lines = file.content.split("\n")`:

```
── BEFORE (main)
  Pattern drift: src/handlers/payments.ts uses http_client while 5/6 files use orm   [confidence 1]

── AFTER (fix/orm-mislabel-87)
  no codedna-pattern findings
```

## The measurement

Population: 11 repositories, 4 language mixes, roughly 4,000 discovered files. Four Vibestack repos plus seven third-party OSS repos already vendored under `eval/discrimination/.cache/`. Run with the cache disabled:

```bash
VIBEDRIFT_DISABLE_CACHE=1 npx tsx src/cli/index.ts "$R" --local-only --no-cache --format json
```

| repo | composite | `codedna-pattern` | of which name `orm` | `drift-architectural_consistency` |
| TanStack-query | 82.5 → 82.5 | 14 → **0** | 14 → **0** | 10 → 10 |
| trpc-trpc | 94.8 → 95.1 | 33 → 19 | 33 → **10** | 1 → 1 |
| honojs-hono | 70.8 → 70.9 | 5 → **0** | 5 → **0** | 0 → 0 |
| fastify-fastify | 88.0 → 88.2 | 5 → **0** | 4 → **0** | 0 → 0 |
| vibedrift-public | 87.0 → 86.8 | 1 → **0** | 1 → **0** | 6 → 6 |
| vibedrift-landing-page | 90.1 → 90.1 | 3 → 3 | 3 → 2 | 8 → 8 |
| vibe-drift-api | 75.0 → 75.0 | 1 → 1 | 1 → 1 | 3 → 3 |
| vibelang | 74.6 → 74.6 | 0 → 0 | 0 → 0 | 1 → 1 |
| expressjs-express, sindresorhus-ky, vuejs-core | unchanged | 0 → 0 | 0 → 0 | 0 → 0 |
| **total** | | **62 → 23** | **61 → 13** | **29 → 29** |

Three things worth reading off that table.

**Recall is preserved where an ORM genuinely exists.** trpc really does use Prisma. Counting signals rather than findings, it keeps exactly **20 of 20** `ORM import/usage` signals after the fix, and 0 Ent false positives. An earlier draft of commit 2 dropped all 20, because `src/server/routers/` is a `routers` directory and my first anchored pattern only accepted `routes`. That is why `routers?` is in the alternation, and why there is a test pinning it.

**No collateral findings were lost.** `drift-architectural_consistency` totals are identical at 29 before and after. Commit 3 can in principle shrink a directory group below `minGroupSize` and silently delete unrelated data-access findings, so this column is the check for that, and it did not happen.

**The composite barely moves, and that is the point.** The largest change is +0.3 on trpc; `vibedrift-public` moves −0.2. The case for this fix is the precision of a user-visible sentence, not the number.

## What this deliberately does not fix

The remaining 13 ORM-naming findings come from two *other* ORM regexes in the same file (`/\.Where\(/i` matching numpy's `np.where(`, `/\.create\(.*{/` matching `stripe.customers.create({`). That is a separate false-positive class with a different fix, and folding it in would have made this measurement uninterpretable. Same for the one surviving Path B case: `routes.findOne(pattern)` passes an identifier rather than a route-path literal, so it is indistinguishable from `Model.findOne(id)` without receiver or import knowledge. It is pinned in the golden corpus with a comment saying so.

I measured the two gates that *would* provide that knowledge before rejecting them: requiring a known ORM token keeps 2 of 8 real ORM idioms, requiring an ORM-ish chain root keeps 4 of 8. The first-argument discriminator shipped here keeps 10 of 10. A manifest gate is not expressible at all, because both entry points take only `(content, path)`.

## Regression proof

I inverted each change and watched the tests fail:

- Restore `|ent\.` in the alternation → **20 of 60** fail in `pattern-classifier.test.ts`
- Restore the unanchored path gate → **8 of 60** fail (exactly the eight rejection cases)
- Remove the `(?!['"\`]\/)` lookahead → **3 fail** in `classifier-golden.test.ts` (`mountChiRoutes`, `mountFiberRoutes`, `mountCustomRouter`)

`test/unit/codedna/pattern-classifier.test.ts` is new; there were previously **zero** tests anywhere referencing this module. It adds 60 tests: 15 true positives across six ORM libraries, 13 identifiers that must no longer match, boundary cases for aliases, string literals, comments and casing, 18 path-gate cases in both directions, and a finding-level test reproducing the reported symptom.

## Blast radius

```
$ command grep -rn "classifyPatterns|patternFindings|isHandlerOrServiceFile" src/
src/codedna/index.ts
src/codedna/pattern-classifier.ts

$ command grep -rn "classifyDataAccessLabel|hasDataAccessPattern" src/
src/drift/architectural-contradiction.ts
src/session/finding-anchor.ts
src/tools-core/tools/validate-change.ts
```

Commit 3 touches all three of its consumers at once by editing the shared regex, which preserves the `classifyDataAccessLabel` ↔ `hasDataAccessPattern` symmetry the module docstring calls load-bearing.

## Version bumps

`BASELINE_VERSION` 3 → 4 in commit 3. The `data_access` vote is persisted in the MCP baseline and read by `validate_change` and the session inline check, so a warm `~/.vibedrift/baseline-cache` would otherwise keep serving the old label and the fix would not reproduce for existing users.

No `Analyzer.version` bump: this touches Code DNA and a drift detector, neither of which is keyed by the findings cache. No `SCORING_VERSION` bump: no scoring math, constant, factor or gate threshold changed. Flagging one judgment call for you rather than deciding it myself — the fix does move stored composites by up to 0.3, which makes historical scores marginally non-comparable in the way a scoring change would. Say the word if you want it bumped.

## Calibration

`npm run calibrate` run before and after. The measured rows are **byte-identical**, and the one enforced gate holds:

```
✓ security floor precision gate: 1.000 >= 0.95
```

The `architectural_consistency` row sits at recall 0.00, which pre-dates this branch: the saved baseline I diffed against is dated 2026-07-16. `npm run calibrate:monotonic` is not required here (no scoring or weight change) and is red on `main` for unrelated reasons.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Refactor
- [ ] Test
- [ ] Chore / infra

## Checklist

- [x] `README.md` updated if a flag, command, or feature changed — n/a, no flag, command or feature changed

Suite: 191 files, 2583 tests, all passing.

## Related issues

Closes #87

---

Investigated and implemented by an AI assistant following `docs/AI_CONTRIBUTOR_GUIDE.md`, including the investigator/refuter passes. The refuter overturned several numbers from the first pass, notably that a bare `\bent\.` boundary is not a sufficient fix.

### DIFF
--- a/src/codedna/pattern-classifier.ts
+++ b/src/codedna/pattern-classifier.ts
@@ -48,7 +48,25 @@ const SIGNAL_DEFS: SignalDef[] = [
   { pattern: "raw_sql", regex: /\.query\s*\(\s*[`'"]/i, label: "inline SQL query string" },
 
   // ORM
-  { pattern: "orm", regex: /(?:gorm|prisma|sequelize|typeorm|sqlalchemy|django\.db|ent\.)\.?/i, label: "ORM import/usage" },
+  { pattern: "orm", regex: /(?:gorm|prisma|sequelize|typeorm|sqlalchemy|django\.db)\.?/i, label: "ORM import/usage" },
+  // Ent (entgo.io) is split out of the alternation above, anchored AND case-sensitive.
+  // A bare `ent\.` had no left boundary, so it matched every identifier ending in the
+  // letters "ent" followed by a dot: content., fullContent., client., component.,
+  // current., event., document., agent., environment., argument., management. That
+  // single alternative produced 125 of 125 "ORM import/usage" signals across five
+  // repos that declare no ORM at all.
+  //
+  // A left \b alone is NOT sufficient: it still fires on a local variable named `ent`
+  // (ent.ty, ent.entitled, ent.capture_types). Real Ent usage is a package selector
+  // followed by an exported Go identifier (ent.Client, ent.NewClient, ent.Open,
+  // ent.Tx, ent.IsNotFound, []*ent.User), whereas a local `ent` is followed by a
+  // lowercase field. The [A-Z] discriminator separates them.
+  //
+  // The /i flag is deliberately omitted: with it, [A-Z] also matches lowercase and
+  // the discriminator becomes a no-op. Known and accepted recall gap: a package
+  // alias (myent.Client, some_ent.Client) is not recognised. The generated Ent
+  // package is named `ent` by default.
+  { pattern: "orm", regex: /\bent\.[A-Z]/, label: "Ent ORM usage" },
   { pattern: "orm", regex: /\.Find\(\s*&|\.Create\(\s*&|\.Save\(\s*&|\.Where\(/i, label: "ORM method call (Go)" },
   { pattern: "orm", regex: /\.findOne\(|\.findAll\(|\.create\(.*{|\.update\(.*{|\.destroy\(/i, label: "ORM method call (JS)" },
   { pattern: "orm", regex: /objects\.(?:filter|get|create|all)\(/i, label: "Django ORM call" },
--- a/src/codedna/pattern-classifier.ts
+++ b/src/codedna/pattern-classifier.ts
@@ -80,9 +80,23 @@ const SIGNAL_DEFS: SignalDef[] = [
   { pattern: "http_client", regex: /fetch\(|axios\.|requests\.(?:get|post)/i, label: "HTTP fetch/axios/requests" },
 ];
 
-// Files that are likely "handler" or "service" files (where pattern drift matters)
+// Files that are likely "handler" or "service" files (where pattern drift matters).
+//
+// SEGMENT match, not substring. Unanchored, "route" admitted router.go,
+// autorouter.ts, itty-router.js and this repo's own src/drift/route-extractors/,
+// and "api" admitted therapist.ts, capital.ts and rapid.ts. Across seven repos,
+// 344 of 499 gated files entered on a substring rather than a path segment, and
+// a file that enters the classifier gets labelled by whatever signal fires first.
+//
+// `routers?` is included deliberately: a directory or file named `router`/
+// `routers` is real routing code and holds real data access (trpc's Prisma
+// usage lives in src/server/routers/). Only the SUBSTRING matches are dropped,
+// so autorouter.ts, itty-router.js and route-extractors/ no longer qualify
+// while routers/post.ts still does.
+//
+// This mirrors the anchored form CONTEXT_PRIORS already uses just below.
 function isHandlerOrServiceFile(path: string): boolean {
-  return /(?:handler|controller|service|route|endpoint|api|resource)/i.test(path);
+  return /(?:^|\/)(?:handlers?|controllers?|services?|routes?|routers?|endpoints?|api|resources?)(?:[/.]|$)/i.test(path);
 }
 
 // ─── Bayesian context priors ─────────────────────────────────────────
--- a/src/core/baseline.ts
+++ b/src/core/baseline.ts
@@ -31,7 +31,7 @@ import type { IntentHint } from "../intent/types.js";
 
 const CACHE_DIR = join(vibedriftHome(), "baseline-cache");
 /** Bump when vote logic / detector set / signature format changes (invalidates all caches). */
-export const BASELINE_VERSION = 3;
+export const BASELINE_VERSION = 4;
 
 export interface CategoryVote {
   driftCategory: DriftCategory;
--- a/src/drift/architectural-contradiction.ts
+++ b/src/drift/architectural-contradiction.ts
@@ -55,8 +55,23 @@ function detectDataAccess(file: DriftFile): { pattern: DataAccessPattern; eviden
     results.push({ pattern: "raw_sql", evidence: sqlEvidence });
   }
 
-  // ORM
-  const ormPatterns = /\.(?:Where|Find|Create|Save|Delete|First|Preload|findOne|findMany|findAll|objects\.filter)\s*\(/g;
+  // ORM. Route registrations share these verb names with ORM CRUD calls:
+  // router.Create("/users", h), r.Delete("/orders/{id}", h.Del),
+  // app.Delete("/users/:id", deleteUser). The discriminator is the FIRST
+  // ARGUMENT: a route path is a string literal beginning with "/", and no ORM
+  // idiom measured here (gorm, prisma, sequelize, typeorm, django, ent,
+  // mongoose) passes one in that position.
+  //
+  // A receiver or import gate would be stronger, but this module has no ORM
+  // vocabulary and no manifest access — classifyDataAccessLabel and
+  // hasDataAccessPattern take only (content, path) and are the MCP
+  // validate_change and session-resolve contracts. Measured alternatives:
+  // requiring a known ORM token keeps 2 of 8 real ORM idioms, requiring an
+  // ORM-ish chain root keeps 4 of 8. Argument shape keeps 10 of 10.
+  //
+  // The lookahead is zero-width, so extractEvidence still advances and still
+  // records line numbers from match.index.
+  const ormPatterns = /\.(?:Where|Find|Create|Save|Delete|First|Preload|findOne|findMany|findAll|objects\.filter)\s*\(\s*(?!['"`]\/)/g;
   const ormEvidence = extractEvidence(c, ormPatterns);
   if (ormEvidence.length > 0) results.push({ pattern: "orm", evidence: ormEvidence });
 
--- a/test/fixtures/classifier-bodies.txt
+++ b/test/fixtures/classifier-bodies.txt
@@ -391,3 +391,42 @@ export async function loadProfile(userId: string) {
   await repo.put(userId, res);
   return { ...res, extra };
 }
+=== mountChiRoutes synthetic ===
+func mountOrderRoutes(r chi.Router, h *OrderHandler) {
+  r.Get("/orders", h.List)
+  r.Create("/orders", h.New)
+  r.Delete("/orders/{id}", h.Del)
+}
+=== mountFiberRoutes synthetic ===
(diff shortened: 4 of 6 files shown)


BrandonGarate177 (approved):
I ran the full loop (npm ci && npm run build && npm test && lint && typecheck), looks all clean, 191 files / 2583 tests, matches what you posted exactly.

Looking through the contributor guide I inverted each of the three commits one at a time and confirmed the tests actually fail, instead of just trusting the regression-proof section. All three matched exactly:

Reverted the Ent regex → 20/60 fail, as claimed
Reverted the path-segment gate → 8/60 fail, as claimed
Removed the lookahead → exactly mountChiRoutes, mountFiberRoutes, mountCustomRouter fail, as claimed
Also cloned a few real repos (fastify, express, ky), and ran the actual built CLI before/after. fastify went from 6 findings (5 ORM) down to 0 on my snapshot, a little different from your 5→0/4→0, probably just because fastify's moved a few commits since you measured, but the before/after delta on my exact snapshot matches the effect you're describing, composite score moved +0.2 both times. Express and ky both confirmed unchanged at 0→0. And npm run calibrate came back byte-identical between main and this branch (minus the timestamp), security floor gate still holds.

Everything checks out, sweet! I'm approving.

The PR description and the golden test comment both say the routes.findOne(pattern) case is "tracked in todo.md", but it isn't in there. Might just not have made it into this branch yet, but wanted to flag it since the PR points to it as the paper trail.

skhan75:
Yes `routes.findOne(pattern)` is tracked in my local todo.md file where I park items to work on  (basically a backlog of items in the order of priority). I'll push this in a separate PR

PR #93, opened 2026-08-05
fix: score clones pasted with their signature

### LINKED ISSUE #81: [bug] validate_change/find_similar_function score exact-logic duplicates as unrelated when the query includes the function signature
### What happened

`validate_change` and `find_similar_function` take a `body` parameter documented as "the function body (source)" / "the proposed/changed function body (source)" — both descriptions read naturally as "paste the function." When a caller does that (signature line included, as any agent copy-pasting a function would), the tools can miss an exact-logic duplicate of an existing function entirely.

Root cause: the baseline's duplicate index (`minhashIndex` in `src/core/baseline.ts:206-214`) is built from `extractAllFunctions(...).rawBody`, which is **body-only** — it excludes the `function name(...) {` declaration line. But the query path (`src/codedna/find-similar-to-body.ts:findSimilarToBody`, called from `src/tools-core/tools/validate-change.ts` and `find-similar-function.ts`) tokenizes whatever text the caller passes, with no equivalent stripping. If the caller includes the signature, the query's token stream carries extra tokens the index never had, which deflates the LCS/Dice similarity score (`src/codedna/minhash.ts:lcsSimilarity`) — badly for short functions, where the signature is a large fraction of the total tokens.

### Steps to reproduce

1. Build the baseline for this repo (`vibedrift .` once, or call any MCP tool that lazily builds it).
2. Note the existing function `src/session/nudge.ts:33-35`:
   ```ts
   export function isNewInteractiveSource(source: string | undefined): boolean {
     return source === "startup" || source === "clear";
   }
   ```
3. Call `validate_change` with a verbatim clone under a different name, **including the signature** (the natural way to call it):
   ```ts
   export function isStartupOrClear(source: string | undefined): boolean {
     return source === "startup" || source === "clear";
   }
   ```
   `rootDir`: repo root. `targetPath`: any new file, e.g. `src/session/scratch.ts`.
4. Observe the result.

### Expected behavior

`duplicateOf` should include `isNewInteractiveSource` — this is a textbook copy-paste-rename duplicate.

### Actual behavior

```json
{"status":"ok","ok":true,"conflicts":[],"duplicateOf":[],"referenceFiles":[],"confidence":"high"}
```
No duplicate detected. Confirmed via direct token comparison that the self-match (identical function, signature included) scores **0.6976** similarity against the index — well under `validate-change.ts`'s `DUPLICATE_THRESHOLD = 0.8` (`src/tools-core/tools/validate-change.ts:29`) — versus **1.0** when the same body is queried without its signature line.

### Environment

- VibeDrift version: 0.19.1
- Node version: v22.22.0
- OS: macOS (darwin arm64)

### Additional context

Possible fixes: (a) run the caller's `body` through the same extraction step used to build the index (strip to braces-inner content) before tokenizing the query, so both sides are on equal footing regardless of what the caller pastes; or (b) if body-only input is intentional, update both tools' parameter descriptions to say so explicitly, since the current wording invites exactly the input that breaks detection.

This isn't just a discovery-tool gap — `validateChangeAgainstBaseline`'s duplicate check is the same machinery the Drift Sessions re-check uses (`signalPresent`'s `redundancy` case in `src/session/finding-anchor.ts`) to decide whether a flagged clone is genuinely gone. That internal path passes body-only text on both sides (self-consistent), so it isn't affected the same way — but it means the in-session and the MCP-tool paths currently give different guarantees on the same detector, which is worth being explicit about even if only the MCP-facing side needs the fix.

### LINKED ISSUE #86: [bug] Drift Session redundancy finding falsely resolves when a flagged clone is wrapped in a class AND edited by one token
## What happened

This confirms a defect the maintainer already documented himself in `todo.md` after an adversarial pass on PR #77 ("redundancy, wrap plus one token") — still present, unfixed, on current `main`. Filing this to attach a concrete, automated repro so it can become a real regression test rather than a prose backlog item.

A redundancy finding's anchor holds the flagged construct's exact normalized token sequence. A clone moved verbatim into a class correctly stays open (token containment holds). But change even one identifier while wrapping, and containment breaks — the re-check falls through to a whole-file duplicate query that's structurally blind to class/object-literal methods (`extractAllFunctions` only indexes top-level `function`/`const` forms). Result: a near-verbatim clone that is still functionally present resolves as if it had been removed.

## Steps to reproduce

1. Baseline contains `src/lib/backoff.ts` with `exponentialBackoff(attempt)`.
2. Raise a redundancy finding by introducing a byte-identical clone elsewhere (e.g. `src/util.ts`).
3. "Fix" it by wrapping the clone in a class **and** renaming one local variable (e.g. `jitter` → `noise`) — not a genuine fix, just a disguise.
4. Re-check the file.

## Expected behavior

The finding should stay open: the duplicated logic is still there, just wrapped and cosmetically altered.

## Actual behavior

`recheckFile` reports the finding resolved.

## Environment

- VibeDrift version: 0.19.1
- Node version: v22.22.0
- OS: macOS (darwin arm64)

## Additional context

Maintainer's own suggested fix (from `todo.md`): a similarity test (minhash the anchor tokens against sliding windows) rather than exact token containment, or the deeper fix of widening `extractAllFunctions` to index class/object-literal methods (noted there as a non-local change affecting scoring across the whole product).

### Repro script (copy-paste runnable, mirrors the project's own `test/unit/session/outcome-anchors.test.ts` fixtures/helpers)

```ts
import { mkdtempSync, mkdirSync, rmSync, writeFileSync, realpathSync } from "node:fs";
import { tmpdir } from "node:os";
import { join } from "node:path";
import { buildBaseline, type RepoDriftBaseline } from "../../src/core/baseline.js"; // adjust path
import { runEditChecks } from "../../src/session/check.js";
import { recheckFile, type OpenFinding } from "../../src/session/outcomes.js";

const HELPER_BODY = `export function exponentialBackoff(attempt: number): number {
  const base = 250;
  const cap = 30_000;
  const jitter = Math.random() * 100;
  return Math.min(cap, base * 2 ** attempt) + jitter;
}`;

async function main() {
  const repo = realpathSync(mkdtempSync(join(tmpdir(), "vd-repro-repo-")));
  const sessionsDir = realpathSync(mkdtempSync(join(tmpdir(), "vd-repro-sessions-")));
  mkdirSync(join(repo, "src", "lib"), { recursive: true });
  writeFileSync(join(repo, "src", "lib", "backoff.ts"), `${HELPER_BODY}\n`);
  const baseline: RepoDriftBaseline = await buildBaseline(repo);

  const out = await runEditChecks({
    rootDir: repo, projectHash: "feedfacefeedface", sessionId: "s1",
    sessionsDir, file: join(repo, "src/util.ts"), body: HELPER_BODY,
    loadBaselineFor: async () => baseline,
  });
  const open: OpenFinding[] = out.flags
    .filter((f) => f.detail.category === "redundancy")
    .map((f) => ({ findingId: f.findingId!, file: f.detail.file!, category: f.detail.category!, anchor: out.anchors[f.findingId!] }));

  const wrappedPlusToken = `export class Backoff {
  exponentialBackoff(attempt: number): number {
    const base = 250;
    const cap = 30_000;
    const noise = Math.random() * 100;
    return Math.min(cap, base * 2 ** attempt) + noise;
  }
}`;

  const result = recheckFile(baseline, "src/util.ts", wrappedPlusToken, open);
  console.log("Resolved (should be []):", JSON.stringify(result.resolved));

  rmSync(repo, { recursive: true, force: true });
  rmSync(sessionsDir, { recursive: true, force: true });
}

main();
```

BrandonGarate177:
## Summary

The duplicate index only stores the *body* of a function since the declaration line gets stripped out (`extractAllFunctions(...).rawBody`). But nothing on the query side strips it back off. 

So if you paste a whole function, signature and all, you're comparing against tokens the index entry never had. For short functions this isn't even a soft penalty: `lcsSimilarity`'s length-ratio gate can slam straight to a hard `0` before it runs any real comparison, since the signature alone can more than double the token count against a tiny body-only entry.

Fix: `findSimilarToBody` now scores the query two ways, as-is, and (when it can find a leading signature to peel off) with that stripped too, keeping whichever score is higher per index entry. 

It doesn't try to be smart about "does this look like a signature", it just tries every `{` in a signature-sized window and picks whichever one's balanced match lands closest to the end of the text. It recurses too, so a class wrapping a method gets peeled in two passes, with a guard so it doesn't mistake a trailing `try`/`catch` for another signature layer.

It's safe by construction: since we're taking the max of two scores, a bad strip just loses to the raw score. It can only push a score up toward what a body-only query would've gotten, never invent a match that isn't real.

**Side effect:** this also fixes #86's reported repro. The session's redundancy re-check goes through `validateChangeAgainstBaseline`, which is literally the same `validateChange` this PR is fixing (`tools-core/index.ts:26` — a re-export, not a lookalike). A class-wrapped clone with one changed identifier stops falsely "resolving." 

That only holds when the wrapped clone is basically the whole file. If it's sitting in a bigger file with unrelated code after it, the false-resolve is still there. That's a problem for another PR, tracked separately against #86.

## Type of change

- [x] Bug fix

## The measurement

Ran every function in this repo's own `src/` (1026 of them) against its own self-clone, pasted with its full signature, before and after:

| | Before | After |
| --- | --- | --- |
| Median similarity | 0.752 | **1.000** |
| Under 0.8 (`validate_change`'s threshold) | 766 (74.7%) | **1 (0.1%)** |
| Under 0.6 (`find_similar_function`'s threshold) | 88 (8.6%) | **0 (0.0%)** |
| Exact hard zero | 46 (4.5%) | **0 (0.0%)** |

One straggler left (`estimateScoreAfterFixes` in `src/scoring/engine.ts`, 0.689) — it's got a JSDoc comment sitting inside its return-type object literal, so no brace lands near the end of the text and the stripper correctly bails instead of guessing wrong. Still clears `find_similar_function`'s 0.6 bar, just not `validate_change`'s stricter 0.8.

Scan script's scratch, not committed (matches the project's own bar-4 workflow) — to redo it: walk `src/**/*.ts` through `extractAllFunctions`, build a `SimIndexEntry[]` from each `rawBody`, then re-query every function's declaration+`rawBody` through `findSimilarToBody` at `threshold: 0`.

## The regression proof

Stashed just the three source files (left the tests in place) and reran the affected suites — 13 of 63 failed against the old code, including the new short-function hard-zero case and the #86 class-wrap case. Popped the stash, all 63 green again.

## The blast radius

`findSimilarToBody` only has two callers and both are in this diff (its signature's untouched, still returns `SimMatch[]`):
```
$ grep -rn "findSimilarToBody" src/
src/tools-core/tools/validate-change.ts:197
src/tools-core/tools/find-similar-function.ts:55
```
Anything calling `validateChangeAgainstBaseline` gets the improvement for free, no code changes needed on their end (it's the same underlying function, and the fix can only raise scores, never lower them):
```
$ grep -rn "validateChangeAgainstBaseline" src/
src/session/detect.ts:104          (finding-raise decomposition path)
src/session/finding-anchor.ts:175  (redundancy re-check -- the #86 connection)
```
The MCP layer (`src/mcp/tools/find-similar-function.ts`, `src/mcp/tools/validate-change.ts`) just re-exports the tools-core functions above, so nothing to touch there.

## Checklist

- [ ] `README.md` updated — n/a, no flag/command/feature surface changed
- [x] Copy uses "Vibe Drift Score" (never "debt score" or "quality score") — n/a, no score-facing copy touched

## Related issues

Closes #81
Related to #86 (closes the reported repro; the general multi-function-file case is tracked separately)

### DIFF
--- a/src/codedna/find-similar-to-body.ts
+++ b/src/codedna/find-similar-to-body.ts
@@ -4,8 +4,125 @@
  * directly (buildSignature → tokens, then exact LCS verify) rather than
  * sampler.ts's band helper, which caps at SIMILARITY_BAND_HIGH (0.80) to
  * EXCLUDE obvious clones — here we want the full range, including exact dups.
+ *
+ * Index/query asymmetry (issue #81): every `SimIndexEntry` is built from
+ * `extractAllFunctions(...).rawBody` (`core/baseline.ts`), which is BODY-ONLY
+ * -- it excludes the function's own declaration line. A caller who pastes a
+ * whole function (signature included, the natural thing to do) gets a query
+ * token stream the index entry never had, which can deflate the similarity
+ * score below either consuming threshold. For short functions this isn't a
+ * gradual deflation -- `lcsSimilarity`'s length-ratio gate can return a hard
+ * 0 before any LCS runs at all, because the signature alone can more than
+ * double the token count relative to a small body-only entry.
+ *
+ * Fix: score the query BOTH as given and, when a leading signature can be
+ * heuristically stripped, with that stripped too, and keep the higher of the
+ * two per index entry. This is monotone -- it can only raise a score toward
+ * what a body-only query would have found, never invent a match that
+ * wouldn't otherwise exist -- so a wrong strip (e.g. on a body that legitimately
+ * opens with a nested block) never makes things worse than not stripping at
+ * all; it just loses to the raw candidate in the max.
  */
-import { buildSignature, lcsSimilarity } from "./minhash.js";
+import { buildSignature, lcsSimilarity, type SimilaritySignature } from "./minhash.js";
+
+/** How far into the text a plausible signature-ending brace/colon must sit to
+ *  be treated as one. Real TypeScript signatures with inline object-type
+ *  parameters or multi-line generic constraints routinely run past 200 chars
+ *  (measured against this repo's own src/), so this is deliberately generous.
+ *  Safe to widen further: the "closest to the end of the text" selection in
+ *  `stripLeadingSignature` is what actually guarantees correctness, not this
+ *  cutoff -- it only bounds how far into a long paste we bother looking. */
+const SIGNATURE_ZONE = 600;
+
+/** Bare keywords that open a brace block without being a declaration --
+ *  `try {`, `catch (e) {`, `if (x) {`, etc. A block like this can still be
+ *  the LAST statement in a real body, which makes its own closing brace land
+ *  suspiciously close to the text's end -- exactly the signal the "closest
+ *  to the end" rule is looking for. Without this guard, recursing into an
+ *  already-correctly-stripped body can mistake its own trailing try/catch
+ *  for another signature layer and strip away real code. */
+const CONTROL_FLOW_KEYWORDS = new Set(["if", "for", "while", "switch", "catch", "try", "else", "do", "finally"]);
+
+/** Is the brace at `braceIdx` immediately preceded (mod an optional
+ *  parenthesized clause, e.g. `catch (e)`) by a bare control-flow keyword? */
+function precededByControlFlow(text: string, braceIdx: number): boolean {
+  const window = text.slice(Math.max(0, braceIdx - 60), braceIdx).replace(/\s+$/, "");
+  const withoutParen = window.replace(/\([^()]*\)\s*$/, "").replace(/\s+$/, "");
+  const words = withoutParen.split(/\s+/);
+  return CONTROL_FLOW_KEYWORDS.has(words[words.length - 1]);
+}
+
+/**
+ * If `text` looks like it opens with a function declaration, return just the
+ * body past that declaration (mirroring the shared extractor's rawBody
+ * convention: opening `{` excluded, closing `}` included; Python: dedented
+ * block after the `:` line). Returns null when no plausible split point
+ * exists, so the caller falls back to scoring the raw text as-is.
+ *
+ * Deliberately language-agnostic: `find_similar_function` has no target file
+ * at all (the function doesn't exist yet), so there's nothing reliable to
+ * detect a language from. Also deliberately doesn't try to recognize a
+ * signature by shape (e.g. `name(` or `name<T>(`) -- generic type parameters,
+ * multi-line param lists, and a return type that itself contains a brace
+ * (`(): { x: number } {`) all defeat a simple "does the prefix look like a
+ * signature" regex. Instead: try every `{` within the signature zone as a
+ * candidate body-open, and keep whichever one's balanced match consumes
+ * closest to the end of the text -- the real body brace is the one whose
+ * matching `}` is (near) the last character, because everything after an
+ * EARLIER brace (like a return-type object literal) is more function left to
+ * go, so its own match closes well short of the end.
+ */
+export function stripLeadingSignature(text: string): string | null {
+  const trimmedLen = text.replace(/\s+$/, "").length;
+  let best: string | null = null;
+  let bestGap = Infinity;
+
+  let searchFrom = 0;
+  for (let braceIdx = text.indexOf("{", searchFrom); braceIdx !== -1 && braceIdx < SIGNATURE_ZONE; braceIdx = text.indexOf("{", searchFrom)) {
+    searchFrom = braceIdx + 1;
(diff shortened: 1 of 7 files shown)


skhan75 (approved):
this looks solid. Nice desription

Issue #82, opened 2026-07-31
[bug] Short-function duplicate detection has no reliable floor — a real collision slips past both the MCP tools and the official analyzer's existing safeguard


BrandonGarate177:
## What happened

MinHash+LCS duplicate detection normalizes every local identifier to `ID0`, `ID1`, ... and every string/number literal to `STR`/`NUM` before comparing token streams (`normalizeTokens` in `src/codedna/minhash.ts`). For short functions this can erase everything that distinguishes two genuinely unrelated functions, so two predicates with the same *shape* but different *meaning* collapse to an identical token stream and score as a 100% duplicate.

This has two independently-confirmed manifestations:

**1. The MCP tools (`find_similar_function`, `validate_change`) — real repo functions, no floor at all.** `src/codedna/find-similar-to-body.ts` has zero length floor. Two real, unrelated functions in this repo — `isNewInteractiveSource` (`src/session/nudge.ts:33-35`, session-start source check) and `isPaidPlan` (`src/auth/plan.ts:11-13`, account plan-tier check) — normalize to the byte-identical 15-token array `["return","ID0","=","=","=","STR","|","|","ID0","=","=","=","STR",";","}"]` and score 100% similarity through the actual MCP tool.

**2. The scan-time `duplicates` analyzer — has a floor, but it's off-by-one.** `src/analyzers/duplicates.ts` guards against short bodies with `if (body.length < 20) continue;` and `if (sig.tokens.length < 15) continue;` (lines 79, 82) — proof the team already recognizes short bodies as unreliable. But the token-count guard is `< 15`, not `<= 15`, so a pair at exactly 15 tokens passes through untouched. Demonstrated with a synthetic pair chosen specifically to avoid #88's separate extraction bug (plain JS-style declarations, no return-type annotation, so `duplicates.ts`'s own regex actually extracts them) — **executed directly against the real `duplicatesAnalyzer.analyze()`**, not just traced:

```
isCancelled(status) { return status === "off" || status === "no"; }   // src/session/gate-a.ts
isDeclined(answer)  { return answer === "off" || answer === "no"; }   // src/auth/gate-b.ts
```

Real output from `duplicatesAnalyzer.analyze()`:
```json
{
  "analyzerId": "duplicates",
  "severity": "warning",
  "confidence": 0.75,
  "message": "1 pair(s) of duplicate/near-duplicate functions detected",
  "locations": [
[...]

PR #94, opened 2026-08-05
fix: keep redundancy findings open for disguised clones (#86) (#84)

### LINKED ISSUE #84: [bug] Drift Session finding resolves on an unrelated hunk when the post-edit file read fails
## What happened

When `readFile(checkAbsFile)` throws during re-check (e.g. the file is briefly unreadable), `hook-entry.ts` falls back to the edit's own `body`. For a `Write` tool call that's the whole file — fine. For an `Edit` tool call, `body` is only the hunk (`new_string`). If that hunk is small and unrelated to the flagged pattern, the re-check sees "the pattern is absent from (what it's treating as) the whole file" and resolves the finding — even though the real file on disk is untouched.

## Steps to reproduce

1. Flag a `.then()` chain via a `Write` PostToolUse event on `src/report.ts`.
2. Make the file unreadable (`chmod 000 src/report.ts`) to force `readFile` to throw on the next check.
3. Fire an `Edit` PostToolUse event on the same file with a small, unrelated hunk (e.g. `old_string`/`new_string` that just adds a comment).
4. Restore file permissions and inspect the ledger and the real file content.

## Expected behavior

On a read failure, the re-check should either skip resolution entirely (leave the finding open — consistent with every *other* fail-open path in this module) or explicitly mark the outcome as inconclusive rather than resolving on a fragment.

## Actual behavior

A `resolve` event is appended to the ledger, while the real on-disk file content (verified directly after the test, permissions restored) still contains the original flagged `.then()` chain, completely untouched.

## Environment

- VibeDrift version: 0.19.1
- Node version: v22.22.0
- OS: macOS (darwin arm64)

## Additional context

Maintainer's own stated fix (from `todo.md`): "skip the re-check entirely on read failure." Not yet implemented as of current `main`.

### Repro script (copy-paste runnable, exercises the real `hook-entry.ts` process exactly like `test/integration/session-hook.test.ts`)

```ts
import { spawnSync } from "node:child_process";
import { mkdtempSync, mkdirSync, readdirSync, readFileSync, writeFileSync, chmodSync, rmSync, realpathSync } from "node:fs";
import { tmpdir } from "node:os";
import { join } from "node:path";

const REPO_ROOT = process.cwd(); // run from repo root
const ENTRY = join(REPO_ROOT, "src", "session", "hook-entry.ts");
const BUILDER = join(REPO_ROOT, "test", "helpers", "session-build-baseline.ts");
const TSX = join(REPO_ROOT, "node_modules", ".bin", "tsx");

function runHook(home: string, payload: unknown) {
  return spawnSync(TSX, [ENTRY], {
    input: JSON.stringify(payload), encoding: "utf8",
    env: { ...process.env, HOME: home, USERPROFILE: home, VIBEDRIFT_HOOK_DEBUG: "" },
    timeout: 30_000,
  });
}
function tmp(prefix: string) { return realpathSync(mkdtempSync(join(tmpdir(), prefix))); }
function ledgerLines(home: string, sid: string) {
  const sessions = join(home, ".vibedrift", "sessions");
  const hashDir = readdirSync(sessions)[0];
  return readFileSync(join(sessions, hashDir, `${sid}.jsonl`), "utf8").trim().split("\n").map((l) => JSON.parse(l));
}

const home = tmp("vd-repro-home-");
const repo = tmp("vd-repro-repo-");
mkdirSync(join(repo, ".git"));
mkdirSync(join(repo, "src"), { recursive: true });
writeFileSync(join(repo, "CLAUDE.md"), "- Async: use async/await throughout. No .then() chains.\n");
for (const n of ["a", "b", "c"]) {
  writeFileSync(join(repo, "src", `${n}.ts`), `export async function ${n}(){ return await fetch("/${n}"); }\n`);
}
spawnSync(TSX, [BUILDER, repo], { encoding: "utf8", env: { ...process.env, HOME: home, USERPROFILE: home }, timeout: 60_000 });

const reportPath = join(repo, "src", "report.ts");
const thenBody = ["export function r() {", '  return fetch("/r")', "    .then((x) => x.json())", "    .then((j) => j.data);", "}"].join("\n");
writeFileSync(reportPath, thenBody + "\n");
runHook(home, {
  session_id: "hunk-1", cwd: repo, hook_event_name: "PostToolUse", tool_name: "Write",
  tool_input: { file_path: reportPath, content: thenBody },
});

chmodSync(reportPath, 0o000); // force readFile to throw on the next check

runHook(home, {
  session_id: "hunk-1", cwd: repo, hook_event_name: "PostToolUse", tool_name: "Edit",
  tool_input: { file_path: reportPath, old_string: "export function r()", new_string: "// unrelated\nexport function r()" },
});

chmodSync(reportPath, 0o644);
const events = ledgerLines(home, "hunk-1");
const resolved = events.some((e) => e.type === "resolve");
const stillFlagged = readFileSync(reportPath, "utf8").includes(".then(");
console.log("resolve event fired:", resolved, "| flagged .then() still on disk:", stillFlagged);
console.log(resolved && stillFlagged ? "BUG CONFIRMED" : "not reproduced");

rmSync(repo, { recursive: true, force: true });
rmSync(home, { recursive: true, force: true });
```

### LINKED ISSUE #86: [bug] Drift Session redundancy finding falsely resolves when a flagged clone is wrapped in a class AND edited by one token

BrandonGarate177:
## Summary

There are two ways the Drift Session re-check can falsely **resolve** a finding whose flagged code is still on disk. These are the two residual resolve defects tracked together in `todo.md`, batched here as one session re-check change. Both live in the same resolution path, and both fail in the dangerous direction: clearing a finding the user never fixed.

### Fix 1: disguised clones (#86)

A redundancy finding could falsely resolve when you disguise a flagged clone and change even one token. Move the clone into a class or object-literal method, or rename the function and tweak one identifier or numeral, and the re-check reports it "resolved" even though the duplicated logic is right there.

The cause is a query asymmetry in the re-check's missing-symbol path (`src/session/outcomes.ts`, the `body === undefined` branch).

It first tries exact token containment, which one changed token breaks. Then it falls back to a whole-file `validateChangeAgainstBaseline` query. That query runs through `extractAllFunctions`, which only indexes top-level `function`/`const` forms, so it is structurally blind to class and object-literal methods.

The wrapped clone is invisible to it, the score dilutes below the duplicate threshold, and the finding clears as if the code were gone. The causal predicate is not the class wrap. It is *"the anchored symbol is no longer findable by name AND at least one token changed."* A plain rename plus one token, no class anywhere, false-resolves too. A fix aimed only at class wrapping would miss half of it.

The fix (`src/session/finding-anchor.ts`): the redundancy presence test now ORs **raw-token shingle containment** of the anchor's own tokens with the whole-file query, so the finding stays open when either check says the clone is still there:

```ts
case "redundancy": {
  if (anchor.kind !== "function") return true;
  return (
    tokenContainment(anchor.tokens, body) >= REDUNDANCY_CONTAINMENT || // 0.6
    validateChangeAgainstBaseline(baseline, relFile, body).duplicateOf.length > 0
  );
}
```

Three things matter about how it works:
- **Raw tokens, not normalized.** `normalizeTokens` (minhash) renumbers identifier slots by first occurrence, so a normalized subsequence embedded in a larger file renumbers and never matches. `tokenizeBody` keeps identifiers literal, which is what survives a wrap.
- **Containment over `|anchor|`, not Jaccard.** A large surrounding file cannot dilute it, and 5-token shingles (matching the minhash window) mean a one-token edit only misses the handful of windows that span it. A very short function is the exception, and the query below backstops it.
- **The query is the backstop.** Containment slips on an all-identifier rename, or a body longer than the anchor token cap, where raw tokens change but normalized-token similarity (with #93's signature strip) still scores the clone a match. Containment covers the wrap-plus-one-token disguise the query dilutes away; the query covers the rename the containment misses.

It is not that containment alone is one-sided safe. It is not: it slips on a full rename (see the measurement). The **OR** is one-sided safe by construction. It can only ever keep a finding open more readily than either check alone, never clear one a stricter test would keep. The failure mode it introduces is a stale-open finding, which costs a re-flag. It never causes a false resolve.

### Fix 2: post-edit read failure (#84)

The re-check reads the post-edit file from disk, but it fell back to the edit body when `readFile` threw. For a `Write` that body is the whole file, which is fine. For an `Edit` it is only the hunk, so a small unrelated hunk made the re-check see "pattern absent from the whole file" and resolve a finding whose flagged code was untouched on disk.

The fix (`src/session/hook-entry.ts`): a read failure now skips resolution entirely and leaves findings open. That is fail-open, like the rest of the module, and it matches the fix stated in `todo.md`.

That turned up a latent test bug, flagged by @skhan75 on the issue. The shipped `session-hook.test.ts` resolve test earned its assertion through this exact fallback, because it never wrote the fixed file to disk, so removing the fallback turns it red. I rewrote it to actually write the fixed content the way a real Write or Edit would leave it, and added a regression test that forces the read failure via file deletion (`ENOENT`, which is portable and user-independent where `chmod 000` is ignored by root and Windows) and asserts no `resolve` fires.

## Type of change

- [x] Bug fix (two related session re-check defects)

## The measurement (Fix 1)

I ran every function in this repo's own `src/`, 964 functions with ≥15 anchor tokens, anchored in 178 of the 211 `.ts` files scanned, through a throwaway `tsx` script. Two things measured: how often the containment half keeps a disguised clone open, and how often it fires against an *unrelated* function.

Swept the shingle size K against three disguise strengths, at the 0.6 threshold (containment only):

| K | rename-all-occurrences (worst case) | rename one local (realistic) | change one numeral | false-open on unrelated |
| 3 | 98.3% | 100% | 100% | 27.9% |
| 4 | 92.5% | 100% | 100% | 22.1% |
| **5 (chosen)** | **88.4%** | **100%** | **99.8%** | **18.6%** |

For the actual bug (a disguise plus a *one-token* cosmetic edit) containment alone is 100% (single rename) and 99.8% (numeral) at K=5. Two footnotes on the "realistic" column: the 100% is for changing a *single occurrence*; a real IDE rename of one local, all its occurrences, measures 92%. The "rename-all-occurrences" worst case (88.4%) is renaming one dominant variable everywhere. Renaming *every* identifier is harsher still: containment slips on about 97% of functions, and that is exactly the case the OR'd query holds open (it scores those clones 1.0).

I picked **K=5** deliberately. It gives the best specificity (fewest stale findings) while still catching 100% of the realistic one-token disguises, and it matches the minhash shingle window already used in `src/codedna/minhash.ts`. A smaller K only buys worst-case sensitivity at a real specificity cost, and the OR already backstops that cost with the query.

Two honest caveats on that 18.6% "false-open" number. First, it is the conservative direction. Second, it is a **cross-file** measurement (anchor vs every other function in `src/`), whereas `recheckFile` only tests containment within the same file, so it overstates the real risk; the within-file rate is closer to 9 to 11 percent. `goNamed`⊂`rustNamed` sit in different files and could never keep each other's finding open, for one. Inspecting the top pairs, it is a mix. Some are genuine duplication (the parallel `security-ast-{go,rust,python}` helpers like `inErroredContext`/`nameSegments`, and the near-identical `writeIntentState`/`writeOutcomeState` sidecar writers). Some are artifacts (a small function whose tokens are a subsequence of a larger one, like `clearFooter`⊂`runLiveTape`, or a coincidental token-skeleton overlap between unrelated functions like `pctToken`⊂`grade`). Not "largely genuine," a mix, and in the safe direction regardless. The query half only fires on genuine duplicates, so it adds no spurious opens beyond what the flag path already raises.

Spot values on the issue's own repro shape (anchor = `exponentialBackoff` body); the verdict is the OR of the two columns:

| disguise | containment | query | verdict |
| --- | --- | --- | --- |
| verbatim class wrap | 1.000 | dup | stays open |
| wrap + renamed local (`jitter` to `noise`) | 0.778 | dup | stays open |
| wrap + changed numeral | 0.889 | dup | stays open |
| all-identifier rename | 0.31 | dup | stays open (via the query) |
| genuine removal | 0.000 | none | resolves |

## The regression proof

**Fix 1 (#86):** the tests bind, both ways. Reverting the return to containment alone fails the new all-identifier-rename test (containment 0.31 resolves it) while the OR keeps it open; reverting to the query alone fails the wrap-plus-one-token tests (the query dilutes a wrapped clone below threshold). All 28 tests in `test/unit/session/outcome-anchors.test.ts` pass with the OR, including #93's own additions. New tests live under `describe("issue #86: a disguised clone with a one-token cosmetic edit stays open")`, including the all-rename case, plus a `describe("tokenContainment (raw-token shingle containment)")` unit block.

**Fix 2 (#84):** a repro (a delete-forced read failure) resolves the finding on the old code and leaves it open after the fix, so it binds too. Tests: the rewritten `session-hook.test.ts` resolve test, which now writes the fixed file to disk, plus a new `"does not resolve a finding when the post-edit file read fails (#84)"`.

## The blast radius

Confined. I enumerated every caller over `src/`:

- `signalPresent` (finding-anchor.ts) is called only by `recheckFile` (outcomes.ts, 3 sites).
- `recheckFile` is called only by `hook-entry.ts` (the re-check path). Its signature is unchanged.
- `tokenContainment` is new, called only in `signalPresent` and tests.
- `validateChangeAgainstBaseline` is used by the **flag path** (`session/detect.ts:104`) and is now also OR'd into the re-check in `signalPresent`. #93's signature-strip improvement to it applies to both, and nothing else about it changes.
- Fix 2 is confined to the read-then-recheck block in `hook-entry.ts`. A `readFile` failure now skips resolution instead of falling back to the edit body.

No version bumps needed. `BASELINE_VERSION` (4) and `SCORING_VERSION` (v14) are untouched, because this changes only the session re-check path, not baseline vote logic, the detector set, the anchor/signature format, or any scoring math. It is not an analyzer, so there is no analyzer `version` to bump either.

## Docs

Two docs described the re-check as a single similarity query and were updated here: `docs/handbook/10-drift-sessions.md` and `docs/AI_CONTRIBUTOR_GUIDE.md`. The handbook was rebuilt in the same PR (`docs/handbook/DEVELOPER_HANDBOOK_OSS.html`).

## Checklist

- [x] `npm run lint`, `npm run typecheck`, `npm run build` pass. Affected unit suite green (28/28). Both fixes shown to bind via repros. The full suite's only failures on my machine are pre-existing sandbox filesystem, symlink, and permission env tests, confirmed unrelated by comparing the identical failing set against `main`.
- [x] New behavior has tests (regression tests for both, including the all-rename case, and the resolve test rewritten so it no longer passes via the removed fallback)
- [x] Stale handbook and contributor docs updated, handbook rebuilt in the same PR
- [ ] `README.md` updated: n/a, no flag, command, or feature surface changed
- [x] This change improves how accurately VibeDrift measures drift (kills two false-resolve classes in Drift Sessions)
- [x] Copy uses "Vibe Drift Score": n/a, no score-facing copy touched

## Verification note

The session-hook *integration* suite spawns `tsx` and could not run in my Windows dev sandbox. The harness invokes the extensionless `.bin/tsx`, which `spawnSync` cannot launch, so it returns `status: null` for all of them, before and after this change. Both fixes were verified by runnable repros through `npx tsx` and by the in-process unit suites. CI (linux) runs the integration tests.

## Related issues

Closes #86, the general multi-function-file case that #93 explicitly deferred (#93 fixed #86's reported repro only when the wrapped clone was basically the whole file).
Closes #84, the sibling resolve defect (`todo.md` bullet (b)), batched here per the note that these should land as one session re-check change. Same cluster as #81 and #82.


BrandonGarate177:
Updated pr with fix for #84

skhan75:
can we rebase this?

src/session/finding-anchor.ts
```diff
@@ -161,18 +197,17 @@ export function signalPresent(
...
+      // forms, so a clone moved into a class or object-literal method is
+      // invisible to the query: it dilutes below the duplicate threshold and
+      // goes quiet on code that never left (#86). Containment over the anchor's
+      // own raw tokens survives a wrap AND a cosmetic one-token edit, and is
+      // one-sided safe — it only ever keeps a finding open more readily, never
+      // clears one a stricter test would have kept. A file anchor has no single
+      // construct to test, so it stays open.
       if (anchor.kind !== "function") return true;
-      return validateChangeAgainstBaseline(baseline, relFile, body).duplicateOf.length > 0;
+      return tokenContainment(anchor.tokens, body) >= REDUNDANCY_CONTAINMENT;
```
skhan75: the containment check should be OR'd with the old query, not replace it. on its own it creates new false resolves the old path caught.

quick example: rename every variable in the clone, like an IDE "rename symbol" pass:

```ts
const floorMs = 250;                 // was base
const ceilMs = 30_000;               // was cap
const wobble = Math.random() * 100;  // was jitter
return Math.min(ceilMs, floorMs * 2 ** turn) + wobble;
```

containment drops to 0.31, under 0.6, so this resolves. but renames are invisible to normalized tokens, so on current main (with #93's signature strip) the old query scores this exact file 1.0 and keeps it open. I measured it across src/: with an all-identifier rename, ~97% of functions slip containment but the query still catches them.

suggested, keeps everything this PR fixes and loses nothing:

```ts
return tokenContainment(anchor.tokens, body) >= REDUNDANCY_CONTAINMENT ||
  validateChangeAgainstBaseline(baseline, relFile, body).duplicateOf.length > 0;
```

the OR version is the one that's actually one-sided safe: it can only keep more findings open than either check alone.

src/session/finding-anchor.ts
```diff
@@ -101,6 +100,43 @@ export function tokensContainedIn(tokens: string[] | undefined, content: string)
   return haystack.includes(` ${tokens.join(" ")} `);
 }
 
+/** Sliding-window size for containment shingles — 5 tokens, matching the
+ *  similarity path's shingling (minhash.ts). Small enough that even a short
+ *  function yields many overlapping windows, so a one-token edit only misses
+ *  the handful of windows that span it. */
+const CONTAINMENT_SHINGLE = 5;
```
skhan75, line 110: "only misses the handful of windows that span it" breaks down for short functions. a 5-token anchor has exactly one window, so any single edit takes containment straight to 0 and the finding resolves.

example, raised through the real hook path: flag a clone of

```ts
export function five(a){ return a + 1; }
```

then change the numeral to `a + 2`. containment is 0.000 and it resolves. the old body query kept this open at similarity 1.0, numerals are normalized away. so the one-token disguise this PR is about comes back at small sizes. the OR fix from my other comment covers this too.

src/session/finding-anchor.ts
```diff
@@ -101,6 +100,43 @@ export function tokensContainedIn(tokens: string[] | undefined, content: string)
...
+/**
+ * What fraction of the anchor's RAW-token shingles are still present in
+ * `content` (0..1)? RAW, not normalized: `normalizeTokens` (minhash) renumbers
+ * identifier slots by first occurrence, so a normalized subsequence embedded in
+ * a larger file renumbers and never matches — a dead end recorded on #86.
+ * Containment over |anchor| (not Jaccard), so a large surrounding file cannot
+ * dilute it, and it survives a wrap plus a cosmetic one-token edit. Zero for an
+ * empty/absent anchor: no evidence is not evidence of presence.
+ */
+export function tokenContainment(tokens: string[] | undefined, content: string): number {
```
skhan75, line 134: two more shapes where containment reads low on code that's still sitting on disk. both stayed open under the old query, both resolve here:

1. anchor tokens are capped at 400, so a long clone only gets tested on its first ~80 lines. delete just those from a 150-line clone and containment reads 0.49 with 110 lines still there verbatim.
2. shingles go into a Set, so a line repeated 12 times collapses to a few unique shingles. rename two vars in the small distinct tail and containment reads 0.44 while the repeated bulk is untouched (a multiset version reads 0.78).

both covered by OR-ing the old query back in.

src/session/finding-anchor.ts
```diff
@@ -101,6 +100,43 @@ export function tokensContainedIn(tokens: string[] | undefined, content: string)
   return haystack.includes(` ${tokens.join(" ")} `);
 }
 
+/** Sliding-window size for containment shingles — 5 tokens, matching the
+ *  similarity path's shingling (minhash.ts). Small enough that even a short
+ *  function yields many overlapping windows, so a one-token edit only misses
+ *  the handful of windows that span it. */
+const CONTAINMENT_SHINGLE = 5;
+
+/** Presence threshold for the redundancy re-check. skhan75 measured this on
```
skhan75: cite the issue here instead of my handle, "measured on #86", we don't name accounts in source comments. and once the OR change is in, the "one-sided safe" wording above the return needs the same fix: containment alone isn't one-sided safe (see the rename example), containment OR the old query is.

skhan75 (commented):
ran everything end to end on macOS including the integration suite you couldn't run on Windows: full gate green, both inversion proofs reproduce exactly as described, the spot table matches to the third decimal, and 964 is exact. fix 2 is good as is. nice measurement discipline again.

one ask before anything else: please rebase on main. #93 merged today after this branch was cut. the merge itself is clean (I tested it), but this PR currently deletes the query #93 just improved on this exact path, which is what makes the OR version in my line comments matter.

few footnotes on the measurement section, all in the safe direction: the 18.6% false-open number is per-anchor ("at least one hit anywhere in src/") and includes same-file pairs (clearFooter sits nested inside runLiveTape), the within-file rate is more like 9-11%. "964 across 211 files": 211 is every .ts file scanned, the anchors live in 178. and the "rename one local (realistic)" row is 100% only when a single occurrence changes, a real IDE rename of one local measures 92%.

also two docs go stale with this change, please touch them up in the same PR: docs/handbook/10-drift-sessions.md:121 (re-check still described as a similarity query) and docs/AI_CONTRIBUTOR_GUIDE.md:596 (raise and re-check sharing one path). nit: both commit subjects are 60+ chars, we keep them near 50.

BrandonGarate177:
Rebased on main (picks up #93) and pushed. What changed:

- **OR'd the query back in.** The redundancy check is now `containment >= 0.6 || validateChangeAgainstBaseline(...)`, not containment alone. Your rename example was right: containment slips on a full identifier rename (only ~3% survive), and #93's signature-stripped query is what holds those open. Added a regression test for that case; it binds both ways (containment-only fails it, query-only fails the wrap-plus-one-token cases).


- **Comments fixed.** Cite #86 instead of my handle, and the "one-sided safe" line above the return now says what you said: containment alone is not one-sided safe, the OR is. Also fixed the short-function caveat on `CONTAINMENT_SHINGLE`.


- **Docs.** Updated `handbook/10-drift-sessions.md` and `AI_CONTRIBUTOR_GUIDE.md` (both described the re-check as a single similarity query) and rebuilt the handbook.


- **Measurement footnotes** in the description corrected: anchors live in 178 of 211 files, the "realistic rename" 100% is a single-occurrence change (a real IDE rename is 92%), and 18.6% false-open is per-anchor cross-file (within-file is 9 to 11 percent).


- Commit subjects trimmed under 50.

Fix 2 (#84) is unchanged.

skhan75 (approved):

Issue #88, opened 2026-07-31
[bug] \`duplicates\` analyzer extraction regex skips every TypeScript function with a return-type annotation


BrandonGarate177:
## What happened

`src/analyzers/duplicates.ts` uses its own function-extraction regex instead of the shared `extractFunctionsFromFile` in `src/codedna/function-extractor.ts`:

```ts
/(?:export\s+)?(?:async\s+)?function\s+(\w+)\s*\([^)]*\)\s*\{/g
```

That pattern requires `)` to be immediately followed by `{`. Typed TypeScript functions write `): ReturnType {` (or `): Promise<T> {`, `): boolean {`, etc.), so the regex matches **zero times** against ordinary TS source.

The shared extractor already handles this: `findBodyOpenBrace` / `skipReturnType` in `function-extractor.ts` skip an optional return-type annotation before locating the body `{`. `semantic-duplication`, Code DNA, and the MCP baseline all go through that path. Only the registered `duplicates` analyzer still uses the weaker regex.

Practical effect: on a TypeScript codebase, `duplicates` under-reports (often to silence) even when near-identical typed functions exist across files. Near-dup signal then depends entirely on `drift-semantic_duplication` / Code DNA / MCP — three other implementations of the same idea that *do* extract the bodies. The handbook (`docs/handbook/04` / Ch 07) still describes `duplicates` as a thin wrapper over the shared MinHash pipeline; the extraction gap is undocumented.

This is independent of the short-body floor bug (separate draft / upstream #82): even a long, clearly duplicated typed function is invisible to this analyzer.

## Steps to reproduce

1. Create two files with byte-identical typed function bodies, e.g.:
   ```ts
   // src/a.ts
   export function clamp01(n: number): number {
     if (n < 0) return 0;
     if (n > 1) return 1;
     return n;
   }

   // src/b.ts
   export function clampOne(n: number): number {
     if (n < 0) return 0;
     if (n > 1) return 1;
     return n;
   }
   ```
2. Run the `duplicates` analyzer (or `vibedrift . --format json` and filter `analyzerId === "duplicates"`).
3. Optionally strip the `: number` return types and re-run.

Alternatively, exercise the regex in isolation against current source:

```js
const dupRe = /(?:export\s+)?(?:async\s+)?function\s+(\w+)\s*\([^)]*\)\s*\{/g;
const withRet = 'export function isPaidPlan(plan?: Plan | null): boolean {\n  return plan === "pro";\n}';
[...]

PR #95, opened 2026-08-05
refactor: dedup escapeRegex + cli confirm prompt

BrandonGarate177:
## Summary

Two generic helpers were hand-copied byte-for-byte across the tree. This lifts the verified-identical copies into one home each. Behavior-preserving, no logic change.

- **`escapeRegex`** (the standard "escape RegExp metacharacters" one-liner, `str.replace(/[.*+?^${}()|[\]\\]/g, "\\$&")`) had **three** identical copies: `codedna/deviation-heuristics.ts`, `analyzers/dead-code.ts`, and `codedna/semantic-fingerprint.ts` (the third differs only in the parameter name `s` vs `str` — identical body). → new `src/core/regex.ts`.
- **A y/N stdin prompt** was duplicated as `askConfirm` in `cli/commands/enable.ts` and `askConsent` in `cli/commands/watch-session.ts` — same body, different name. → new `src/cli/prompt.ts` exporting `askConfirm`. Both call sites already used it as a default (`options.confirm ?? …`), so the swap is a one-liner each, and `import readline from "readline"` drops out of both files since it was only used inside that function.

Only genuinely-identical, generic helpers with no calibration surface. 

I explicitly did **not** touch two same-named false friends, because they're different functions:
- `core/file-filter.ts` has its own `escapeRegex(ch)` — a single-character *glob* escaper (`/[.+^${}()|\\]/`, no `*?[]`), different semantics.
- `cli/commands/init.ts` has `ask(rl, question)` — returns the raw answer string, not a y/N boolean, and takes an existing interface.

## Type of change

- [x] Refactor (no behavior change)

## The measurement

Found by grouping every function in `src/**/*.ts` by exact normalized token stream (`extractAllFunctions` + `tokenizeBody`), reporting groups that span ≥2 files. `escapeRegex` (×3) and the y/N prompt (×2) are byte-identical; both fall out of that grouping. No third implementation of either exists in `src/` (the two false friends above are structurally different and were left alone).

## The regression proof

Behavior-preserving by construction — the extracted functions are character-identical to the copies they replace (only the prompt's name is unified to `askConfirm`). Verified:

- `npm run typecheck`, `npm run lint`, `npm run build` — clean.
- Targeted suites for every touched module — **80/80 pass**: `dead-code`, `deviation`, `semantic-fingerprint`, `semantic-duplication`, `enable`, `watch-session`.
- Full suite: the failing set is unchanged versus clean `main` (all pre-existing environment tests — session-hook/mcp-tee/git-metadata/etc.; the ±1 count is a known flaky env test). None of the failures are in the touched modules.

## The blast radius

- `escapeRegex` → `src/core/regex.ts`, imported by exactly the 3 former definers (`deviation-heuristics.ts`, `dead-code.ts`, `semantic-fingerprint.ts`). Enumerated every `escapeRegex` reference in `src/`; the only other is `file-filter.ts`'s unrelated glob escaper, left untouched.
- `askConfirm` → `src/cli/prompt.ts`, imported by `enable.ts` and `watch-session.ts`. `readline` import removed from both (no other use). `init.ts`/`feedback.ts` have their own distinct prompts, left untouched.
- No version bumps: no analyzer logic, scoring math, baseline vote/signature, or detector set changed — pure helper relocation. (`escapeRegex` lives inside `dead-code` and `semantic-fingerprint`, but their *behavior* is byte-identical, so cached outputs are unaffected.)

## One review note

`git` renders `src/codedna/semantic-fingerprint.ts` as a **binary** diff ("Binary files differ"). That's pre-existing and unrelated to this change: the file uses raw `\0` (NUL) bytes as sentinels around literal placeholders (`` `\0LIT${idx}\0` ``), and git flags any file with a NUL as binary. HEAD's copy has the same 2 NUL bytes. My change there is exactly the −48-byte removal of the local `escapeRegex` function plus one import line — visible via `git show --text` or by reading the file.

## Checklist

- [x] `npm run lint`, `npm run typecheck`, `npm run build` pass; touched-module tests green (80/80). Full-suite failures are pre-existing env tests, unrelated.
- [x] No new behavior (pure refactor) — covered by the existing tests of each touched module
- [ ] `README.md` updated — n/a, no flag/command/feature surface changed
- [x] Improves how confidently VibeDrift measures drift — indirectly: three independent `escapeRegex` copies (two inside analyzers) can no longer silently diverge
- [x] Copy uses "Vibe Drift Score" — n/a, no score-facing copy touched

## Related

- Surfaced by VibeDrift's own semantic-duplication detector on this repo (see #88 / `todo.md`, which cite `escapeRegex` and `nameSegments` as dogfood duplicates). This PR is the source cleanup, not the #88 detector-scoping decision.


src/cli/commands/enable.ts
```diff
@@ -16,7 +16,7 @@
  * supported agent is detected; the baseline stays lazy (first scan builds it).
  */
 
-import readline from "readline";
+import { askConfirm } from "../prompt.js";
```
skhan75, line 19: Why are we importing a `ts` file as `js`?

skhan75 (changes requested):

src/cli/commands/watch-session.ts
```diff
@@ -9,7 +9,7 @@
  * and it is off the hook's hot path.
  */
 
-import readline from "readline";
+import { askConfirm } from "../prompt.js";
```
skhan75, line 12: same as above

skhan75 (changes requested):
One minor comment, rest all looks good to me

BrandonGarate177:
I went and checked the repo before changing it, and I think the `.js` here is actually the house convention rather than a mistake (do NOT quote me on this) 

This is a TypeScript ESM project, so a relative import points at the `.ts` source using the emitted `.js` specifier. It is not a one-off in this file:

- 643 of 643 relative module imports in `src/` end in `.js`. Zero end in `.ts`, zero are extensionless (I scanned `src/**/*.ts`, excluding comment lines).
- `src/` has no `.js` source files at all, so every one of those `.js` imports already resolves to a `.ts` file.
- The other imports in this same `enable.ts` do exactly this: `../../session/repo.js`, `../../session/install.js`, `../../session/activation.js`, `../../core/vibedrift-home.js`.
- `tsconfig` is `moduleResolution: "bundler"`, so extensionless would still compile, but the codebase has standardized on `.js`.

So `import { askConfirm } from "../prompt.js"` (pointing at the new `src/cli/prompt.ts`) matches every existing import. Changing it to `../prompt` or `../prompt.ts` would make it the only relative import in `src/` that does not use `.js`.

skhan75:
you're right, checked it myself. it's 676/676 on current main now (the count grew after the recent merges), zero real extensionless imports, the few grep hits for it are comment lines in import-graph-ast.ts. and tsup bundles dist, so the specifier never even reaches users.

for anyone landing on this thread later: typescript never rewrites import paths when it compiles. so in an ESM package you write the path as it will exist after build (.js) and the checker maps it back to the .ts source. looks odd, is correct, and it stays working if we ever move off bundler resolution to nodenext or per-file emit, which extensionless wouldn't.

so keeping the .js convention as is. resolving these threads, and appreciate you checking the census before changing anything.

skhan75 (approved):

PR #96, opened 2026-08-05
refactor: dedup identical security-AST helpers

BrandonGarate177:
## Summary

Putting this up as a **gander, not a strong recommendation** cause the comments made me think the modularization here was on purpose for separation of concerns. 

While auditing duplication I noticed the per-language security-AST analyzers (`security-ast-{go,rust,python}.ts`) carry a few byte-identical helpers, and I thought it might make sense to share them. The three verified-identical ones are:

- **`nameSegments`** — the identifier → lowercase-segments splitter. In **all three** files, byte-identical.
- **`inErroredContext`** — the parent-walk parse-error guard. In **go + rust**, byte-identical.
- **`namedChildrenOf`** — non-null named children. Was `goNamed`/`rustNamed` (identical bodies); unified under a neutral name (`namedChildrenOf`, not `named`, to avoid shadowing go's local `const named` variables).

This lifts them into `src/drift/security-ast-common.ts` and imports them from the three analyzers.

## What I deliberately did NOT touch

Two pairs *look* like duplicates (same token structure) but differ in the literals that matter, the per-language grammar node types. So unifying them blindly would break one language:

- `goPrunedDescendants` skips `func_literal`; `rustPrunedDescendants` skips `closure_expression` + `function_item`.
- `collectFunctionDefs` scans `function_definition` (python); `collectRustFunctionDefs` scans `function_item` (rust).

Those stay in their own files. So does everything genuinely language-specific (string-literal decoding, router-name collection, etc.).

## Type of change

- [x] Refactor (no behavior change)

## The measurement

Found by grouping every function in `src/**/*.ts` by exact normalized token stream and reporting cross-file groups. Then I read each candidate's full body to confirm byte-identity before lifting, which is how the two divergent pairs above got caught (they group as "same structure" but their node-type literals differ).

## The regression / identity proof

Behavior-preserving by construction (the extracted bodies are character-identical to the copies they replace; only `goNamed`/`rustNamed` were renamed). Verified:

- `npm run typecheck`, `npm run lint`, `npm run build` — clean; no leftover `goNamed`/`rustNamed`.
- **Security suites: 1018/1018 pass** across 17 files.
- **`npm run calibrate`** (required for security-touching changes): the enforced **`security-floor` precision gate holds at 1.000 ≥ 0.95**, and every posture row is unchanged — `security_posture` and `security_posture_{go,py,rust}` all report `prec 1.00 / recall 1.00 / F1 1.00`. Behavior is identical.
- Full suite: failing set unchanged versus clean `main` (all pre-existing environment tests; the ±1 count is a known flaky env test). No failures in `drift/` or `security`.

## The blast radius

- New `security-ast-common.ts` imported by exactly the three security-AST analyzers. `SyntaxNode` comes from `../core/types.js` (same as its consumers).
- Call-site renames confined to `goNamed`→`namedChildrenOf` (go, 15) and `rustNamed`→`namedChildrenOf` (rust, 7). No other file references these symbols.
- **No version bump needed.** This changes only *where* helpers live, not what they compute — vote logic, the detector set, and the anchor/signature format are untouched, and the calibration confirms identical detection output, so warm `BASELINE_VERSION` caches stay valid. `SCORING_VERSION` is likewise untouched.

## Checklist

- [x] `npm run lint`, `npm run typecheck`, `npm run build` pass; security suites 1018/1018; calibration security-floor gate 1.000 ≥ 0.95
- [x] No new behavior (pure refactor) — covered by the existing security suites + calibration
- [ ] `README.md` updated — n/a
- [x] Improves how confidently VibeDrift measures drift — the Go/Rust/Python copies of these security helpers can no longer silently diverge
- [x] Copy uses "Vibe Drift Score" — n/a, no score-facing copy touched

### DIFF
--- a/src/drift/security-ast-common.ts
+++ b/src/drift/security-ast-common.ts
@@ -1,0 +1,48 @@
+/**
+ * Helpers shared verbatim by the per-language security-AST analyzers
+ * (`security-ast-{go,rust,python}.ts`). Each was a byte-identical copy in two or
+ * three of those files; extracting them here means they cannot silently diverge
+ * — a real hazard in the NEVER-FALSE-BLESS layer. Language-specific helpers
+ * (node-type skip lists, function-def node kinds, string-literal decoding)
+ * deliberately stay in their own files.
+ */
+
+import type { SyntaxNode } from "../core/types.js";
+
+/**
+ * Lowercase segments of an identifier, split on underscore/non-alphanumeric AND
+ * CamelCase boundaries (digits stay attached to their run):
+ *   "get_author_stats" → [get, author, stats]; "JWTBearer" → [jwt, bearer];
+ *   "OAuth2PasswordBearer" → [o, auth2, password, bearer].
+ * Whole-segment matching makes substring blessing structurally impossible.
+ */
+export function nameSegments(name: string): string[] {
+  return name
+    .replace(/([a-z0-9])([A-Z])/g, "$1 $2")
+    .replace(/([A-Z]+)([A-Z][a-z])/g, "$1 $2")
+    .toLowerCase()
+    .split(/[^a-z0-9]+/)
+    .filter((s) => s.length > 0);
+}
+
+/**
+ * True when `node` or any ancestor BELOW `source_file` carries a parse error — a
+ * per-construct surgical skip, never a whole-file fallback. Error recovery can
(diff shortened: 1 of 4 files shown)


skhan75 (approved):
Nice refactor. Much needed

PR #98, opened 2026-08-09
fix: render credits from the live API schema

skhan75:
## Summary

Every signed-in `vibedrift status` prints the literal line **`Deep scans: 1 free + undefined purchased`**, with Free-tier copy even on Pro. Root cause: the client renders `available_purchased` / `available_total` from a credits schema the deployed `GET /account/credits` no longer returns — the live route responds with `{plan, unlimited, deep_scans_this_month, deep_scans_limit, deep_scans_remaining, has_free_deep_scan}` and none of the `available_*` fields. A 200 with a changed shape never throws, so the "skip the line on error" guard never fires and JS interpolates `undefined`.

The same response object misleads two more surfaces: `login`'s `available_total` branch is dead code, and the scan-time **🎁 "1 FREE DEEP SCAN EVERY MONTH"** banner shows on every default Pro scan, because `has_free_deep_scan` means "a deep scan is possible now" (`is_unlimited or remaining > 0`, api route line 221), not "has a Free-tier credit".

Client-side fix, API untouched:

- `CreditsResponse` now matches the live schema; new `isCreditsResponse` structural guard — an unknown shape **hides the line** instead of interpolating (this schema has drifted once already; no line beats a wrong line).
- `status`: `Deep scans: 197 remaining (1/12 used this period)`; free-scan nudge only while the free plan's monthly scan is actually unspent; `0 remaining (12/12 used this period)` + upgrade hint when exhausted.
- `login`: remaining-count hint for paid/topped-up accounts; the free-monthly banner is gated by the new `hasUnspentMonthlyFreeScan`.
- `scan`: the 🎁 banner uses the same gate — Pro users no longer get Free-tier copy on every scan.

## Before / after (real runs against the production API, same Pro account)

| Before — published 0.19.6 | After — this branch |
| --- | --- |
| before | after |
| `Deep scans: 1 free + undefined purchased`<br>`Run 'vibedrift . --deep' to use your free credit.` | `Deep scans: 197 remaining (1/12 used this period)` |

Scan-time banner, same fixture, signed-in Pro: published CLI opens with `🎁  1 FREE DEEP SCAN EVERY MONTH`; this branch prints no banner. (Screenshots committed under `docs/media/`; if inline embeds don't render, the blob links above open directly.)

## Measurement

Population: response shapes the deployed route can emit, plus the legacy pre-migration shape and a malformed control. Harness: a scratch script importing the real exported builders, with the "before" logic copied verbatim from `main:src/cli/commands/status.ts:73-82`.

- Original 8-variant matrix: **before prints literal `undefined` in 4/8 variants; after 0/8**.
- Independent refuter pass extended the population to 10 (fresh-month Pro also hits the `undefined` branch; free-unspent-plus-top-up): **before 5/10, after 0/10**. Legacy and malformed shapes now hide the status line and fall back to the legacy login hint.
- Live check on the real endpoint: `curl` returns HTTP 200 with exactly the six `deep_scans_*`-family keys — no `available_*` keys (raw output in the refuter log).
- `vibedrift usage` (which reads `/account/usage`) already agreed with the API; `status` now agrees with `usage`.

## Regression proof

14 tests in `test/unit/cli/credits-lines.test.ts` (rendering for live/legacy/malformed/unlimited/exhausted/topped-up shapes, no-`undefined` assertions, banner-gate predicate). Inverted the change (`git stash push -- src/`, test kept): **all 14 fail**; restored: 14 pass. Full suite: 194 files / 2634 tests green.

## Blast radius

- `command grep -rn "fetchCredits" src/` → exactly 3 call sites: `status.ts:72`, `login.ts:137`, `scan.ts:131`; all three migrated.
- Dead-field readers on main: `status.ts:76,78,79`, `login.ts:154,156`, plus the interface declaration `auth/api.ts:86-89`. The branch has zero `available_*` references in `src/`.
- Not affected: `usage.ts` (different endpoint), everything non-auth (signed-out runs never fetch credits), `--json`/`--deep` scans (banner path already suppressed).
- Adjacent, out of scope, filed in that repo's backlog: the landing dashboard keeps an independent stale mirror of the old schema (`vibedrift-landing-page/src/lib/dashboard/queries.ts:241-254`); it does not call the endpoint, nothing breaks today.

## Gates

`npm run build` / `npm test` (2634) / `npm run lint` (0 problems) / `npm run typecheck` all pass. Calibration harnesses N/A — no detector, classifier, or scoring change; no analyzer `version`, `SCORING_VERSION`, or `BASELINE_VERSION` bump required (CLI command surfaces only). Verified by an independent refuter pass (verdict: CONFIRMED-WITH-CORRECTIONS; its corrections are folded into the numbers above).

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation
- [ ] Refactor
- [ ] Test
- [ ] Chore / infra

## Checklist

- [x] `README.md` updated if a flag, command, or feature changed (not needed — README does not document the credits line)
- [x] This change improves how accurately or confidently VibeDrift measures drift (or is neutral infra) — neutral infra; it stops the CLI from reporting entitlements that do not exist
- [x] No secrets, credentials, pricing, or internal strategy added (screenshots show the account email and the token prefix `status` is designed to display)

## Related issues

No open GitHub issue — found by the 2026-08-08 release smoke sweep; tracked in `todo.md` (bullet to be removed when this merges).

---
🤖 Opened by Claude Code on Sami's behalf — review and merge decisions are his.


src/auth/api.ts
```diff
@@ -80,18 +80,55 @@ export interface FeedbackResponse {
   received_at: string;
 }
 
+/** Shape of GET /account/credits as the deployed API actually returns it
+ *  (api CreditsResponse): period allowance plus top-up credits, the same
+ *  accounting the /v1/analyze gate enforces. `deep_scans_limit` and
+ *  `deep_scans_remaining` are -1 when `unlimited` is true. */
 export interface CreditsResponse {
-  plan: "free" | "pro" | "enterprise";
+  plan: string;
```
skhan75: For uncle claude --> Why did you remove allowed plan types?

src/auth/api.ts
```diff
@@ -80,18 +80,55 @@ export interface FeedbackResponse {
...
   unlimited: boolean;
-  available_total: number;
```
skhan75, line 86: For uncle Claude --> Why did we remove 86 to 91? We do calculate available total, purchased and others in the backend?

src/auth/api.ts
```diff
@@ -80,18 +80,55 @@ export interface FeedbackResponse {
...
+   *  NOT "has an unspent Free-tier credit". */
   has_free_deep_scan: boolean;
 }
 
+/** Structural guard for CreditsResponse. Rendering must go through this:
+ *  the credits schema has changed shape once already, and a 200 with a
+ *  different shape does not throw — interpolating a missing field printed
+ *  literal "undefined" to every signed-in `status` user. Unknown shape ⇒
+ *  the caller hides the line instead of guessing. */
+export function isCreditsResponse(v: unknown): v is CreditsResponse {
```
skhan75, line 118: For uncle claude --> unknown shouldn't be the type here. Why don't we have a strict type check here? Credits schema shape shouldn't have breaking changes.

src/auth/api.ts
```diff
@@ -80,18 +80,55 @@ export interface FeedbackResponse {
...
+  return (
+    isCreditsResponse(credits) &&
+    !credits.unlimited &&
+    credits.plan === "free" &&
+    credits.deep_scans_this_month === 0 &&
+    credits.has_free_deep_scan
+  );
+}
+
 export class VibeDriftApiError extends Error {
```
skhan75, line 158: For Claude --> We should start using a centralized class for all VibeDrfit exception codes and error handling

src/cli/commands/login.ts
```diff
@@ -172,6 +153,39 @@ async function handleLoginSuccess(
   }
 }
 
+/** Pure renderer for the post-login deep-scan hint. Returns null on any
+ *  shape the structural guard does not recognize so the caller can fall
+ *  back to the legacy hint — the credits schema drifted once already, and
+ *  the free-tier banner used to fire for Pro accounts because
+ *  `has_free_deep_scan` means "can deep scan now", not "has a free credit". */
+export function buildLoginCreditsLines(credits: unknown): string[] | null {
```
skhan75: Shouldn't we have a proper interface for credits, instead of unknown type?

src/cli/commands/login.ts
```diff
@@ -172,6 +153,39 @@ async function handleLoginSuccess(
...
+  if (!isCreditsResponse(credits)) return null;
+  if (credits.unlimited) {
```
skhan75, line 162: what's unlimited?

src/auth/api.ts
skhan75: Done — `plan` is typed again as of e96fee0: `CreditsPlan = "free" | "pro" | "enterprise" | "team" | (string & {})`.

It deliberately isn't the old closed union: the deployed route also emits `team` (account.py's CreditsResponse docstring), which the old union didn't include, and a closed union enforced by the runtime guard would hide the credits line for any plan name this CLI version hasn't heard of — the same schema-drift failure class this PR exists to fix. The open fallback keeps autocomplete and narrowing for the known plans while an unfamiliar plan name stays parseable and renders through the generic path (covered by a new test). Prod today: free=54, pro=2.

*— response drafted by Claude Code at Sami's request*

src/auth/api.ts
skhan75: The backend computes those numbers but never sends them: `get_scan_credits_summary` (api/db.py:511) has zero call sites anywhere in api/ — grep returns only the definition — and the live `GET /account/credits` responds with exactly `{plan, unlimited, deep_scans_this_month, deep_scans_limit, deep_scans_remaining, has_free_deep_scan}` (verified with a real authenticated curl, HTTP 200). Typing fields the wire never carries is precisely what rendered literal `undefined`, so the client interface now mirrors the wire.

If you want the per-source split surfaced in the CLI (e.g. `11 of 12 plan scans left + 186 credits`), the route has to start returning the summary again — that's an API+CLI follow-up I can open; say the word. Left out of this PR on purpose (client-side fix only).

*— response drafted by Claude Code at Sami's request*

src/auth/api.ts
skhan75: Done — restructured in e96fee0 so `unknown` no longer appears at any call site. `fetchCredits` now fetches the payload as untrusted JSON, validates it through the new `parseCreditsResponse` (throws `VibeDriftApiError` on any shape mismatch, which lands in the callers' existing skip/fallback paths), and returns a typed `CreditsResponse`. The builders and `hasUnspentMonthlyFreeScan` all take the proper interface now.

`unknown` survives in exactly one place: the parameter of the boundary guard itself. That's the strict-typing idiom for untrusted network JSON — typing that parameter as `CreditsResponse` would make it a blind cast, and a blind cast (`jsonFetch<CreditsResponse>`) is exactly what shipped the `undefined` bug. On 'the schema shouldn't have breaking changes': agreed as policy, but it did break once already — this validation is what turns the next break into a hidden line plus a throw instead of rendered `undefined`.

*— response drafted by Claude Code at Sami's request*

src/auth/api.ts
skhan75: Agreed on direction — but it touches every command's catch path, so it's scoped as its own focused change rather than a rider on this bugfix. Filed at the top of `todo.md` with a concrete shape: one error module with typed codes (auth / entitlement / schema / network / rate-limit), every API call classifying through it, commands switching on code instead of ad-hoc catches. Not implemented in this PR.

*— response drafted by Claude Code at Sami's request*

src/cli/commands/login.ts
skhan75: Done — `buildLoginCreditsLines(credits: CreditsResponse)` as of e96fee0. Validation moved into `fetchCredits` (throws on unknown shapes), so this builder always receives the proper interface and the legacy-hint fallback rides the throw — same shape as status.

*— response drafted by Claude Code at Sami's request*

src/cli/commands/login.ts
skhan75: It's the server-computed 'this plan has no deep-scan cap' flag: the API maps the plan through `PLAN_DEEP_SCAN_LIMITS` and when that's unlimited (enterprise today) it returns `unlimited: true` with `deep_scans_limit` and `deep_scans_remaining` both -1. This branch renders the 'Run vibedrift . --deep' hint with no numbers because there's nothing to count down.

Done — documented on the `CreditsResponse.unlimited` field itself in e96fee0 (along with every other field, including the -1 sentinel behavior), so the next reader doesn't have to ask.

*— response drafted by Claude Code at Sami's request*

src/auth/api.ts
BrandonGarate177: re-typed as an open union in e96fee0. Verified true. 

`api.ts:88 has CreditsPlan = "free" | "pro" | "enterprise" | "team" | (string & {})`, and there's a test asserting an unseen plan name still parses.

src/auth/api.ts
BrandonGarate177: The live route returns only the `deep_scans_*` family. Client side verified, no way of verifying backend from here. Looks good to me 😄

BrandonGarate177:
## How I went at it

Pulled the branch down, synced my local main to the exact parent commit so `main..pr-98` is just your two commits and nothing else, then went through it, ran it, recompute the numbers from scratch, read the whole diff as one unit, and poke at the gates. 

-----

**The bug is real and I reproduced it.** On main, `status.ts` does `Deep scans: 1 free + ${credits.available_purchased} purchased`, and main's `CreditsResponse` still declares the `available_*` fields. The live route doesn't send those, so it interpolates literal `undefined`. Confirmed straight off the main source.

**The migration is actually complete, not mostly complete.** Zero `available_*` references left anywhere in `src/`. `fetchCredits` has exactly three call sites (`status.ts`, `login.ts`, `scan.ts`) and all three moved over. Nothing left reading the old shape.

**The wire boundary is the right shape.** `isCreditsResponse` type-checks all six live fields, `parseCreditsResponse` throws `VibeDriftApiError` on anything that doesn't match, and `fetchCredits` runs the payload through it before handing back a typed object. Keeping `unknown` only on the guard's own parameter is exactly right. Typing it as `CreditsResponse` would be the blind cast that shipped the bug in the first place. "No line beats a wrong line" is the correct call for a schema that's already drifted once.

**The banner gate is where the Pro-copy fix actually lives.** Reading the builders as one unit, `scan.ts` flips from `has_free_deep_scan && !unlimited` (which Pro satisfies, so Pro kept getting free-tier copy) to `hasUnspentMonthlyFreeScan(credits)`. That's the real fix for the second surface, and the reasoning that `has_free_deep_scan` means "can scan right now" and not "has a free credit" matches what the flag is.

**The tests are honest.** Realistic fixtures captured live (`livePro`, `liveFreeFresh`, `liveExhausted`, `liveUnlimited`, `freeToppedUpSpent`), not two-line toys. And they bind the actual bug with `not.toContain("undefined")` and `not.toContain("free +")` on real rendered output, plus coverage for legacy-rejected, malformed-rejected, unknown-plan-accepted, the `-1` sentinel not leaking, exhausted, and topped-up. These aren't decorative.

**I recomputed the measurement instead of trusting it.** Wrote a throwaway script that runs main's verbatim before-logic and your `buildDeepScanLines` over ten live-schema shapes. After: `0/10` prints `undefined`, matches your claim exactly. Before: I got `6/10` on my ten shapes (you say 5/10 on yours). Not a contradiction, just a different population. The real characterization is "every non-unlimited plan with `has_free_deep_scan: true` renders `undefined`," and that held on every such shape.

**Gates are fine.** `typecheck` and `build` pass here, the credits suite is green. No version bump needed is correct: this is CLI command surface only, no detector, scoring, analyzer, baseline, or signature change, so nothing to bump.


------------
## Notes: 

- the test count is stale, the description says 14 tests in `credits-lines.test.ts` but the file has 17 now. 
- **two slightly different predicates for "show free copy."** `status`'s nudge uses `plan === "free" && deep_scans_this_month === 0`, while `scan` and `login` go through `hasUnspentMonthlyFreeScan` (same thing plus `has_free_deep_scan`). Behaves the same in practice, but routing `status` through the shared gate too would keep it to one definition.
- Description says the bug is tracked in `todo.md` with a bullet to remove on merge. It not there

BrandonGarate177 (approved):

PR #103, opened 2026-08-20
fix: score duplicate volume, not directory count (#102)

### LINKED ISSUE #102: [bug] drift-semantic_duplication never sets dupGroupSize, so near-duplicate volume does not affect the Redundancy score
## What happened

`src/drift/semantic-duplication.ts` emits findings with `countBased: true` but never sets
`dupGroupSize`. The scoring engine has a branch specifically for duplicate detectors that
scores by duplicated-function *fraction*, and it is gated on that field:

```ts
// src/scoring/engine.ts:461
if (functionCount > 0 && findings.some((f) => (f.dupGroupSize ?? 0) > 1)) {
  // ... redundant += (f.dupGroupSize! - 1) * findingMaxWeight(f)
```

The comment directly above it states the intended contract:

> *"Grouped exact/near-duplicate detectors carry `dupGroupSize`: score by the
> DUPLICATED-FUNCTION FRACTION (Σ redundant copies ÷ total functions), which is size-fair AND
> volume-sensitive — **32 identical functions register as ~31 redundant copies no matter how
> they were chunked into findings.**"*

Because `drift-semantic_duplication` does not carry the field, that branch is unreachable for
it. It falls through to the generic count-based branch, `rate = findings.length / functionCount`.
Its findings are **rolled up per directory** (`semantic-duplication.ts`, the `dir`-keyed
aggregation), so `findings.length` is the number of affected *directories* — not the number of
duplicate pairs. The pair count is rendered into the message string and then discarded before
scoring.

**Net effect: 1 near-duplicate pair and 20 near-duplicate pairs in the same directory produce an
identical Redundancy score.**

## Steps to reproduce

Everything needed is inline below — no files to fetch, nothing to clone beyond the repo itself.

```bash
git clone https://github.com/VibeDrift/VibeDrift.git && cd VibeDrift
npm ci && npm run build
# paste the script below to a file, then:
bash repro.sh "$PWD"
```

The generator builds five repos identical in every dimension the scorer normalizes against —
**40 functions, 2 files, 1 directory** — varying only `k`, the number of cross-file
near-duplicate pairs.

The clones are *near*-duplicates by construction: bodies differ only in one string literal.
MinHash tokenization collapses string literals to `STR`, so `drift-semantic_duplication` sees
them as duplicates; `codedna/semantic-fingerprint.ts` preserves literals verbatim, so it does
**not** fire. That isolates the detector under test — confirmed by the `fingerprint` column
reading `0` at every `k`.

> **Reproduction gotcha (this produced a wrong result on my first attempt):** do not place the
> fixtures in a path containing `fixture`, `test`, `spec`, or `mock`. `isNonShippablePath` and
> `computeFileImportanceWeight` substring-match those unanchored, which suppresses the Code DNA
> layer entirely and drops file importance to 0.05. Byte-identical content under
> `…/fixtures/k20` vs `…/plainrepos/k20` scores 90.4 vs 85.2. The script uses `mktemp -d`
> for exactly this reason.

<details>
<summary><b>repro.sh</b> — self-contained, ~70 lines (click to expand)</summary>

```bash
#!/usr/bin/env bash
# Self-contained reproduction. Usage: bash repro.sh /path/to/VibeDrift-checkout
# Fixtures go under mktemp deliberately: isNonShippablePath and
# computeFileImportanceWeight substring-match "fixture"/"test"/"spec"/"mock"
# unanchored, which suppresses the Code DNA layer and masks the result.
set -uo pipefail
REPO="${1:?usage: bash repro.sh /path/to/VibeDrift-checkout}"
CLI="$REPO/dist/cli/index.js"
WORK="$(mktemp -d)"; trap 'rm -rf "$WORK"' EXIT

echo "vibedrift $(node "$CLI" --version)  node $(node --version)  $(uname -sr)"
echo "commit    $(git -C "$REPO" rev-parse --short HEAD)"
echo
echo "dupGroupSize setters under src/:"
command grep -rl "dupGroupSize:" "$REPO"/src --include='*.ts' 2>/dev/null | sed "s|$REPO/|  |" \
  || grep -rl "dupGroupSize:" $(find "$REPO/src" -name '*.ts') 2>/dev/null | sed "s|$REPO/|  |"
echo "in src/drift/semantic-duplication.ts:"
grep -c "dupGroupSize" "$REPO/src/drift/semantic-duplication.ts" 2>/dev/null | sed 's/^/  occurrences: /'
echo

# Fixtures: 40 functions, 2 files, 1 directory. Only near-duplicate pair count varies.
# Clones differ by one string literal -> MinHash collapses literals (semantic-duplication
# fires) but semantic-fingerprint preserves them (codedna-fingerprint does NOT fire),
# isolating the detector under test.
python3 - "$WORK" <<'PY'
import os, sys
PER_FILE = 20
def clone(side, i):
    return f"""export function nearPair{side}{i}(input: string, acc: number): string {{
  const s0 = pairStage{i}_0(acc, input);
  const s1 = pairStage{i}_1(acc, s0);
  const s2 = pairStage{i}_2(acc, s1);
  const s3 = pairStage{i}_3(acc, s2);
  const label = "marker-{side.lower()}-{i}";
  return pairFinish{i}(s3, label);
}}"""
def uniq(side, i):
    n = 3 + (i % 7)
    body = "\n".join(f"  const u{j} = solo{side}{i}_{j}(u{j-1});" if j else
                     f"  const u0 = solo{side}{i}_0(seed);" for j in range(n))
    return f"""export function unique{side}{i}(seed: number): number {{
{body}
  return solo{side}{i}_end(u{n-1});
}}"""
for k in (1, 2, 5, 10, 20):
    root = os.path.join(sys.argv[1], f"k{k:02d}")
    os.makedirs(os.path.join(root, "src", "services"), exist_ok=True)
    open(os.path.join(root, "package.json"), "w").write('{"name":"f","version":"1.0.0","type":"module"}\n')
    a = [clone("Alpha", i) for i in range(k)] + [uniq("A", i) for i in range(PER_FILE - k)]
    b = [clone("Beta",  i) for i in range(k)] + [uniq("B", i) for i in range(PER_FILE - k)]
    for name, fns in (("mod-a.ts", a), ("mod-b.ts", b)):
        open(os.path.join(root, "src", "services", name), "w").write("\n\n".join(fns) + "\n")
print("fixtures: k=1,2,5,10,20 (40 functions, 2 files, 1 directory each)")
PY
echo
printf "%-4s %-11s %-18s %-14s %-13s %s\n" "k" "composite" "Redundancy(health)" "semantic-dup" "fingerprint" "pairs"
printf -- "-----------------------------------------------------------------------------------\n"
for k in 01 02 05 10 20; do
  node "$CLI" "$WORK/k$k" --local-only --format json --no-cache 2>/dev/null \
  | K="$k" python3 -c "
import json,sys,re,os
d=json.load(sys.stdin); r=d['scores']['redundancy']
sd=[f for f in d['findings'] if f['analyzerId']=='drift-semantic_duplication']
fp=[f for f in d['findings'] if f['analyzerId']=='codedna-fingerprint']
p=sum(int(m.group(1)) for f in sd for m in [re.search(r'(\d+) pair',f['message'])] if m)
print(f\"{os.environ['K']:<4} {d['compositeScore']:<11} {str(r['score'])+'/'+str(r['maxScore']):<18} {len(sd):<14} {len(fp):<13} {p}\")
"
done
echo
echo "Redundancy is a HEALTH score: 20/20 = no duplication penalty."
echo "Expected if volume were scored: it DROPS as k rises.  Actual: pinned at 19.7/20."
```

</details>

## Expected behavior

> **Terminology, to avoid an easy misread:** the Redundancy *category score* is a **health**
> score, not a quantity of redundancy. `categoryScore = maxScore × Π(1 − damage)`
> (`engine.ts:548-566`), so **20/20 means no redundancy penalty** and a lower number means more
> duplication was penalized. "Redundancy 19.7/20" therefore reads as *almost none detected*.

**The Redundancy category score should go DOWN as duplicate volume goes UP.** Twenty duplicated
function pairs is a materially worse codebase than one, and the engine already has the machinery
to express that — `codedna-fingerprint` uses it correctly on exact duplicates.

Observed instead: the score is pinned at 19.7/20 across a 20× change in volume, i.e. VibeDrift
rates a codebase with twenty near-duplicate pairs as just as free of redundancy as one with a
single pair.

## Environment

- VibeDrift version (`vibedrift --version`): **0.20.0**
- Node version (`node --version`): **v22.22.0**
- OS: **macOS (Darwin 25.4.0, arm64)**
- Commit: **`7eb89b4`** (tag `v0.20.0`), upstream `main`

## Scan output or logs

Result table (full transcript in the collapsed block below):

```
k     composite   redundancy    semantic-dup     fingerprint    pairs reported
--------------------------------------------------------------------------------
01    90.4        19.7/20          1                0              1
02    90.4        19.7/20          1                0              2
05    90.4        19.7/20          1                0              5
10    90.1        19.7/20          1                0              10
20    90.1        19.7/20          1                0              20
```

The Redundancy health score is flat at **19.7/20** across a 20× change in duplicate volume —
a full-health rating in both cases. Composite moves 0.3 points, and that movement is not
attributable to the duplication.

<details>
<summary>Full run transcript</summary>

```
# VibeDrift issue reproduction transcript
# captured: 2026-08-19T21:18:24Z
# host os : Darwin 25.4.0 arm64
# note    : produced with tee(1); script(1) needs a TTY, unavailable in this shell.
#           Shareable recording:  asciinema rec dup.cast -c './repro.sh /path/to/VibeDrift'
########################################################################

== environment ==
vibedrift : 0.20.0
node      : v22.22.0
os        : Darwin 25.4.0
commit    : 7eb89b4 (v0.20.0)
workdir   : /var/folders/vt/y78zdv8n6md5jgy19jzv1mph0000gn/T/tmp.XswUPSbyjl

== defect in source ==
-- every dupGroupSize setter under src/ --
  src/codedna/semantic-fingerprint.ts
-- does src/drift/semantic-duplication.ts set it? --
  NO — never set

== fixtures: 40 functions, 2 files, 1 directory; ONLY near-duplicate pair count varies ==
  generated k=1,2,5,10,20 (40 functions, 2 files, 1 dir each)

== scan results ==
k     composite   redundancy    semantic-dup     fingerprint    pairs reported
--------------------------------------------------------------------------------
01    90.4        19.7/20          1                0              1
02    90.4        19.7/20          1                0              2
05    90.4        19.7/20          1                0              5
10    90.1        19.7/20          1                0              10
20    90.1        19.7/20          1                0              20

Redundancy is a HEALTH score: 20/20 = no duplication penalty.
Expected if volume were scored: the Redundancy score DROPS as k rises.
Actual: it is pinned at 19.7/20 from 1 pair to 20 pairs -- full health either way.
```

</details>

Source-level confirmation:

```
$ command grep -rln "dupGroupSize:" src/**/*.ts
src/codedna/semantic-fingerprint.ts          # the only setter in the tree

$ command grep -n "dupGroupSize" src/drift/semantic-duplication.ts
                                             # no matches
```

### Mechanism
`countBased: true` causes `driftFindingToFinding` to drop `driftSignal`, routing the finding to
`groupDeviation`'s count branch. That branch computes `findings.length / functionCount`. Because
findings are rolled up per directory, `findings.length` is invariant to pair count.

### Reachability
Not gated. `drift-semantic_duplication` is registered in `createDriftDetectors()`, is drift-kind
in `CATEGORY_CONFIG` (Redundancy), and fires on any default local scan. No flag required. Path
confirmed by running the shipped CLI end-to-end, not by reading.

### User-visible outcome
The Vibe Drift Score under-reports redundancy on the most common shape of AI-generated
duplication — near-identical handlers that differ in a literal or a call. A repo where half of
every module is a near-clone of its sibling scores as effectively clean on that category.

### Magnitude (measured)
At k=20, `functionCount=40`, one `warning` finding, confidence 0.85, importance 1.0:

| | deviation | damage | Redundancy score (health) |
|---|---|---|---|
| current (count branch) | 0.0488 | 0.0166 | **19.7/20** |
| if `dupGroupSize` were set | 0.9643 | 0.3279 | **13.4/20** |

A **6.3-point drop** in the health score on a 20-point category (19.7 → 13.4, i.e. from
"effectively clean" to "visibly penalized"). Computed from the shipped constants
(`COUNT_RATE_SCALE = 0.5`, `DUP_FRACTION_SCALE = 0.15`, `SEVERITY_DAMAGE.warning = 0.4`); the
19.7 figure is confirmed empirically by the scan above, the 13.4 is arithmetic on the same
formula and is **not** an empirical measurement of a patched build.

### Blast radius — narrower than it first appears
- **Exact duplicates are scored correctly.** `codedna-fingerprint` sets `dupGroupSize` and hits
  the fraction branch. Only duplicates in the band "LCS ≥ 0.7 but not literal-identical" are
  unscored. My first attempt at this issue over-claimed that duplicate volume was unscored in
  general; that is false, and the corrected claim is the one above.
- Affects the Redundancy category only. Other four categories unaffected.
- The finding is still *reported* accurately — the pair count is correct in the message and in
  `--format json`. This is a scoring bug, not a detection bug.

### Blast radius — wider
Three detectors set `countBased: true` and therefore share the "count findings, not instances"
branch: `semantic-duplication.ts`, `phantom-scaffolding.ts`, `security-suppression.ts`.
`phantom-scaffolding` also rolls up per directory (`byDir` map), so 2 phantom exports and 50
phantom exports in one directory likewise produce one finding and equal damage. `dupGroupSize`
is duplicate-specific and would not fix that; a general instance-count field would. **I have not
measured the phantom-scaffolding case** — flagging it as a likely sibling, not a verified claim.

### Suggested fix
Set `dupGroupSize` on the finding so the existing fraction branch engages. The natural value is
the count of distinct functions participating in duplication within the rolled-up directory
(the union-find cluster membership already computed for the `clusterSize` severity tier), so
`Σ (dupGroupSize − 1)` approximates redundant copies as the engine comment intends.

Cache/version implications if changed: this alters scoring output, so `SCORING_VERSION`
(currently `"v15"` as of v0.20.0) needs a bump plus a history entry, per `AGENTS.md`.

### What I could not verify
- Whether patching it produces the 13.4 figure. I did not modify and rebuild.
- Whether the correction regresses `npm run calibrate`. The `redundancy` row is report-only
  (only `security-floor` precision is gated), and `calibrate:monotonic` is already red on `main`
  at the 0→10% step, so a clean before/after would need capturing regardless.
- Whether `phantom-scaffolding` and `security-suppression` are affected in practice.
- Any causal claim about tool-assisted sessions producing more duplication. See below — that
  observation is what prompted this investigation, but it is **not** evidence for this bug.

### How this surfaced (observational, not evidence)
Found while analyzing a controlled two-track build experiment (same spec, one track with the
VibeDrift MCP in the agent loop, one without). The tool-assisted track finished with ~8× the
duplicated source lines by jscpd, while VibeDrift scored it 97.7 vs 80.3. Its duplication was
near-duplicate (3 `drift-semantic_duplication` findings covering 19 pairs, zero
`codedna-fingerprint`); the control track's included an exact duplicate that *did* carry
`dupGroupSize` and *was* scaled. That asymmetry is what pointed at this code path. The
experiment is n=1 per track and has a disclosed blinding defect, so it establishes nothing on
its own — the fixture reproduction above is the actual evidence.

skhan75:
Fixes #102.

## The defect

`drift-semantic_duplication` is emitted as `countBased: true` but never carried
`dupGroupSize`, so the scoring engine's duplicate-fraction branch was unreachable
for it. It fell through to the generic count branch, where the divisor is
`findings.length` — and because its findings are rolled up **per directory**, that
is the *directory* count. The pair count was rendered into the message string and
discarded before scoring.

Net effect, reproduced: one near-duplicate pair and twenty in the same directory
scored identically. Redundancy pinned at **19.7/20** across a 20x change in volume.

The root cause is one step deeper than the report states: `DriftFinding` had no
`dupGroupSize` field at all, so the detector could not have set it.

## What changed

Three edits plus the version bump:

1. `src/drift/types.ts` — `DriftFinding` gains `dupGroupSize?: number`.
2. `src/drift/index.ts` — `driftFindingToFinding` carries it through.
3. `src/drift/semantic-duplication.ts` — populated from the union-find the
   detector already builds, so this is arithmetic on existing data.

## The two decisions this required

Neither is a tuning knob, and the issue's implied model gets the first one wrong.

**Copies, not pairs.** `dupGroupSize - 1` means redundant *copies* — that is how
`semantic-fingerprint.ts` sets it and how `engine.ts` consumes it. A cluster of
`m` mutually-duplicate functions is `m(m-1)/2` pairs but only `m-1` redundant
copies; one member is the original. Counting pairs would overstate quadratically,
by up to 2.3x on the worst repo measured.

**Each cluster counted once.** A pair spanning two directories emits a finding in
*both*, so attributing a cluster's copies to every directory it touches would
inflate the total (measured 36% inflation: 321 directory-attributed slots for 236
unique pairs). Each cluster now keeps one canonical member — lowest key, so the
choice is deterministic — and every other member is attributed to the directory
that holds it. Summed across directories that is exactly `m-1` per cluster, once.

A directory holding only cluster originals carries `dupGroupSize: 1` and
contributes no damage. That is correct: the redundancy belongs to whoever holds
the copies.

## Scoring impact

**Scores move. `SCORING_VERSION` v15 → v16, so baselines rebuild and the
first scan after upgrading suppresses its delta.**

Measured against v0.20.0:

| repo | composite | redundancy |
|---|---|---|
| A (Next.js, 37k lines) | 90.2 → 83.1 | 17.7 → 13.8 |
| B (Next.js, 23k lines) | 89.9 → 86.1 | 19.2 → 16.9 |
| C (TypeScript, 81k lines) | 86.2 → 84.1 | 19.7 → 18.3 |
| D (Rust + Python, 55k lines) | 75.3 → 73.3 | 17.2 → 16.3 |
| E (TypeScript extension) | 97.3 → 97.2 | 19.8 → 19.7 |
| F (Python API) | 75.0 → 74.6 | 19.9 → 19.7 |

Repos whose duplicate directories hold a single pair each are unchanged, as are
repos with no near-duplicates. A wider corpus run across all five supported
languages is in progress and its distribution will be posted here before merge.

**Worth stating plainly:** this is a larger movement than the issue's framing
suggests, and it is in the direction of *lower* scores. The detector was
under-penalising real duplication. It is also the second scoring migration in a
short window, so it may be worth batching with other scoring work rather than
shipping alone.

## Testing

Three tests, each verified to fail against `main`:

- redundant-copy count reaches the engine and survives conversion to a `Finding`
- the count scales with duplicate volume, not with directory count
- a cluster spanning two directories is not double-counted

`npm test` 2728 passing, `npm run lint` clean, `npx tsc --noEmit` clean.

## Not in this PR

Two findings from triaging #102 that deserve their own issues:

- **`phantom-scaffolding` has the identical defect** — `countBased: true`, no
  `dupGroupSize`, 30 findings covering 126 phantom exports with the largest
  carrying 30. Left out to keep this reviewable; its severity grading on
  `deadShare` partially mitigates it.
- **An absolute-path leak disables `codedna-fingerprint`.** It calls
  `allNonShippable()` on `SourceFile.path`, which is the *absolute* path, so a
  repo checked out under `/examples/`, `/tests/`, `/spec/` or `/mocks/` silently
  loses exact-duplicate detection. This is what the issue's "reproduction gotcha"
  was really hitting; the functions it names are correctly segment-anchored.

## Still unmeasured

Whether the 236 near-duplicate pairs found across the measured repos are true
positives. This PR changes how much they count, not whether they should count at
all. If detector precision is poor, this makes wrong findings louder.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,33 @@ All notable changes to `@vibedrift/cli` are documented here. The format
 follows Keep-a-Changelog loosely; breaking-shape changes are called out
 explicitly under **Breaking** so CI users can recalibrate.
 
+## [Unreleased]
+
+### Fixed
+
+- **Duplicate volume now affects the Redundancy score.** `drift-semantic_duplication`
+  was emitted as count-based but never carried `dupGroupSize`, so the scoring
+  engine's duplicate-fraction branch was unreachable for it and it fell through
+  to the generic count branch, whose divisor is the number of *findings*. Because
+  its findings are rolled up per directory, one near-duplicate pair and twenty in
+  the same directory produced an identical score. The detector now reports the
+  redundant copies each directory holds, so twenty duplicates register as twenty.
+  Reported in #102.
+
+  Two properties the fix has to preserve, both pinned by tests: it counts
+  redundant *copies* rather than pairs, because a cluster of m mutually-duplicate
+  functions is m(m-1)/2 pairs but only m-1 redundant copies; and a cluster
+  spanning two directories is counted once rather than in both, by attributing
+  each copy to the directory that actually holds it.
+
+### Scoring
+
+**Composite scores move in this release and baselines rebuild.** `SCORING_VERSION`
+advances to v16. Repos with several near-duplicate functions in one directory
+score lower, which is the defect being corrected. Repos whose duplicate
+directories hold a single pair each are unchanged, as are repos with no
+near-duplicates.
+
 ## 0.20.0 — 2026-08-19
 
 **The in-loop checks stop telling you your own conventions are wrong, and the function index finally sees the methods it was blind to.**
--- a/src/drift/index.ts
+++ b/src/drift/index.ts
@@ -323,6 +323,10 @@ export function driftFindingToFinding(d: DriftFinding): Finding {
     // they have no real dominance ratio, so we omit driftSignal and let the engine
     // size-normalize them through its count-based density branch instead of reading
     // a fabricated consistencyScore as a deviation rate.
+    // A count-based DUPLICATE detector still carries volume: `dupGroupSize`
+    // routes it to the engine's duplicate-fraction branch instead of the
+    // generic count branch, whose divisor is finding count (issue #102).
+    ...(d.dupGroupSize !== undefined ? { dupGroupSize: d.dupGroupSize } : {}),
     ...(d.countBased
       ? {}
       : {
--- a/src/drift/semantic-duplication.ts
+++ b/src/drift/semantic-duplication.ts
@@ -160,6 +160,42 @@ export const semanticDuplication: DriftDetector = {
     }
     const sizeOf = (f: ExtractedFunction): number => clusterSize.get(find(fnKey(f))) ?? 1;
 
+    // Redundant copies per directory (issue #102).
+    //
+    // The scoring engine's duplicate-fraction branch reads `dupGroupSize - 1` as
+    // the number of REDUNDANT copies and sums it across a detector's findings.
+    // Two properties have to hold for that sum to mean anything:
+    //
+    //   1. It must count copies, not pairs. A cluster of m mutually-duplicate
+    //      functions is m(m-1)/2 pairs but only m-1 redundant copies — one
+    //      member is the original. Counting pairs would overstate quadratically.
+    //   2. It must not double-count. A pair spanning two directories emits a
+    //      finding in BOTH (see byDir below), so attributing a cluster's copies
+    //      to every directory it touches would inflate the sum.
+    //
+    // Both hold if we pick one canonical member per cluster and attribute every
+    // OTHER member to the directory that actually holds it. Summed over
+    // directories that is exactly (m-1) per cluster, once. The canonical member
+    // is the lowest fnKey so the choice is deterministic — same input, same
+    // attribution, which the repo requires of anything reaching a score.
+    const clusterMembers = new Map<string, string[]>();
+    for (const node of parent.keys()) {
+      const root = find(node);
+      const list = clusterMembers.get(root);
+      if (list) list.push(node);
+      else clusterMembers.set(root, [node]);
+    }
+    const redundantByDir = new Map<string, number>();
+    for (const members of clusterMembers.values()) {
+      if (members.length < 2) continue;
+      // fnKey is `${file}:${name}`; sort by code unit, never localeCompare.
+      const sorted = [...members].sort((a, b) => (a < b ? -1 : a > b ? 1 : 0));
+      for (const member of sorted.slice(1)) {
+        const dir = directoryOf(member.slice(0, member.lastIndexOf(":")));
+        redundantByDir.set(dir, (redundantByDir.get(dir) ?? 0) + 1);
+      }
+    }
+
     // Group pairs by directory — the directory that owns the higher-numbered
     // duplicate is where we attribute the drift.
     const byDir = new Map<string, DuplicatePair[]>();
@@ -234,6 +270,10 @@ export const semanticDuplication: DriftDetector = {
         severity,
         confidence: 0.85,
         countBased: true,
+        // 1 + redundant copies held by THIS directory. Findings whose directory
+        // holds only cluster originals carry 1 and contribute no damage, which
+        // is correct: the redundancy belongs to whoever holds the copies.
+        dupGroupSize: (redundantByDir.get(dir) ?? 0) + 1,
         finding: `${dir}/: ${dirPairs.length} pair(s) of semantically duplicate functions detected (MinHash + LCS verified)`,
         dominantPattern: "unique function bodies",
         dominantCount: 0,
--- a/src/scoring/engine.ts
+++ b/src/scoring/engine.ts
@@ -107,12 +107,23 @@ import {
  *        1 to 34% and duplicate pairs moved -8% to +27%. Repos with no class
  *        or object-literal methods, no property-chain data references and no
  *        `//` inside literals are byte identical.
+ *   v16: duplicate VOLUME reaches the score. `drift-semantic_duplication` was
+ *        emitted as count-based but carried no `dupGroupSize`, so the engine's
+ *        duplicate-fraction branch was unreachable for it and it fell through to
+ *        the generic count branch, whose divisor is the number of FINDINGS.
+ *        Findings are rolled up per directory, so one near-duplicate pair and
+ *        twenty in the same directory scored identically. The detector now
+ *        carries redundant copies per directory, attributing each cluster's
+ *        copies to the directory that holds them so a cluster spanning two
+ *        directories is not counted twice. Repos with several near-duplicates in
+ *        one directory score lower; repos whose duplicate directories hold a
+ *        single pair each are unchanged. Reported in #102.
  *
  * A change here is absorbed silently for users: stored scores are re-aligned
  * where possible and a one-time release-notes notice is shown (see
  * src/core/scoring-notice.ts). Users never see this string.
  */
-export const SCORING_VERSION = "v15";
+export const SCORING_VERSION = "v16";
 
 /** The bundled corpus distribution, typed. Placeholder until the corpus build lands. */
 export const scorePercentiles = scorePercentilesArtifact as ScorePercentiles;
--- a/src/drift/types.ts
+++ b/src/drift/types.ts
@@ -152,6 +152,15 @@ export interface DriftFinding {
    * carried for the report bars, but it does not drive the composite.
    */
   countBased?: boolean;
+  /**
+   * For duplicate detectors: 1 + the number of REDUNDANT copies this finding
+   * accounts for (so `dupGroupSize - 1` is the redundancy it contributes).
+   * Carried into `Finding` so the scoring engine's duplicate-fraction branch
+   * can see duplicate VOLUME; without it a count-based duplicate detector
+   * falls through to the generic count branch, where the divisor is the
+   * number of findings and volume is discarded. See issue #102.
+   */
+  dupGroupSize?: number;
   deviatingFiles: DeviatingFile[];
   /**
    * Up to 3 files that exemplify the dominant pattern — used by fix-prompt
--- a/test/unit/drift/semantic-duplication.test.ts
+++ b/test/unit/drift/semantic-duplication.test.ts
@@ -106,3 +106,67 @@ describe("semantic-duplication detector", () => {
     }
   });
 });
+
+describe("semantic-duplication dupGroupSize (issue #102)", () => {
+  // Bodies that MinHash treats as duplicates: identical structure, differing
+  // only in a string literal (tokenize collapses literals to STR).
+  const clone = (name: string, marker: string) => `export function ${name}(input: string, acc: number): string {
+  const s0 = stageAlpha(acc, input);
+  const s1 = stageBravo(acc, s0);
+  const s2 = stageCharlie(acc, s1);
+  const label = "${marker}";
+  return finish(s2, label);
+}`;
+
+  it("carries redundant-copy count so duplicate VOLUME reaches the scoring engine", () => {
+    // Three mutually-duplicate functions in one directory = 1 original + 2 copies.
+    const ctx = mkCtx([
+      file("src/svc/a.ts", clone("alpha", "m-a")),
+      file("src/svc/b.ts", clone("bravo", "m-b")),
+      file("src/svc/c.ts", clone("charlie", "m-c")),
+    ]);
+    const findings = semanticDuplication.detect(ctx);
+    expect(findings.length).toBeGreaterThan(0);
+
+    // The engine's duplicate-fraction branch is gated on dupGroupSize > 1
+    // (src/scoring/engine.ts). Without it the detector falls through to the
+    // count branch, where findings.length is the DIRECTORY count and duplicate
+    // volume is discarded.
+    const total = findings.reduce((n, f) => n + ((f.dupGroupSize ?? 0) > 1 ? f.dupGroupSize! - 1 : 0), 0);
+    expect(total).toBe(2);
+
+    // and it must survive the conversion into a scoring Finding
+    const converted = findings.map((f) => driftFindingToFinding(f));
+    expect(converted.some((f) => (f.dupGroupSize ?? 0) > 1)).toBe(true);
+  });
+
+  it("scales with duplicate volume rather than with directory count", () => {
+    const two = semanticDuplication.detect(
+      mkCtx([file("src/svc/a.ts", clone("alpha", "m-a")), file("src/svc/b.ts", clone("bravo", "m-b"))]),
+    );
+    const four = semanticDuplication.detect(
+      mkCtx([
+        file("src/svc/a.ts", clone("alpha", "m-a")),
+        file("src/svc/b.ts", clone("bravo", "m-b")),
+        file("src/svc/c.ts", clone("charlie", "m-c")),
+        file("src/svc/d.ts", clone("delta", "m-d")),
+      ]),
+    );
+    const copies = (fs: ReturnType<typeof semanticDuplication.detect>) =>
+      fs.reduce((n, f) => n + Math.max(0, (f.dupGroupSize ?? 1) - 1), 0);
+    expect(copies(two)).toBe(1);
+    expect(copies(four)).toBe(3);
+  });
+
+  it("does not double-count a cluster that spans two directories", () => {
+    // One pair, two directories. The detector emits a finding for EACH
+    // directory, so a naive sum would report two redundant copies for one
+    // duplicated function. Exactly one copy is redundant.
+    const findings = semanticDuplication.detect(
+      mkCtx([file("src/one/a.ts", clone("alpha", "m-a")), file("src/two/b.ts", clone("bravo", "m-b"))]),
+    );
+    expect(findings.length).toBe(2);
+    const total = findings.reduce((n, f) => n + Math.max(0, (f.dupGroupSize ?? 1) - 1), 0);
+    expect(total).toBe(1);
+  });
+});


skhan75:
## Corpus validation is in, and it does not support merging this as written

I said the wider run would post before merge. It has, and it disagrees with the
6-repo sample the PR description was built on.

**212 repositories, five languages, shuffled sample, each scanned with this branch
and with v0.20.0.**

| Language | repos | median | mean | worst |
|---|---|---|---|---|
| Go | 39 | −2.20 | −3.05 | −11.8 |
| JavaScript | 47 | −1.50 | −3.22 | −19.5 |
| Python | 37 | −1.40 | −2.21 | −12.0 |
| **Rust** | 45 | **−3.70** | **−4.50** | **−24.9** |
| TypeScript | 44 | −2.15 | −2.89 | −13.8 |
| **All** | **212** | **−2.10** | **−3.22** | **−24.9** |

**192 of 212 repositories score worse. Only 19 are unchanged. One improves.**
47 drop 5 or more points; 11 drop 10 or more.

For contrast, v0.20.0 shipped with a median of 0.00 and 42% of repos byte-identical.

Worst movers:

```
starship (rust)        96.2 -> 71.3   -24.9
underscore (js)        54.5 -> 35.0   -19.5
htmx (js)              76.4 -> 58.7   -17.7
moment (js)            68.2 -> 52.1   -16.1
dayjs (js)             80.6 -> 66.1   -14.5
sniffnet (rust)        89.5 -> 75.4   -14.1
shadcn-ui (ts)         70.5 -> 56.7   -13.8
open-webui (python)    69.2 -> 57.2   -12.0
```

## Why I think the mechanism is right and the calibration is wrong

Two effects stack, and only one of them is the fix:

1. Duplicate volume now reaches the score. This is the defect being corrected.
2. Routing into the duplicate-fraction branch silently swaps `COUNT_RATE_SCALE`
   (0.5) for `DUP_FRACTION_SCALE` (0.15) — **3.3x harsher** — and replaces
   `groupImportance` (up to 1.5) with a flat 1.0.

The second was never a deliberate choice for this detector. It is inherited from a
branch calibrated for `codedna-fingerprint`, which reports exact clones. This
detector reports *near* duplicates at a 0.7 cosine floor, which is a different and
much looser population, and it is now being scored on the exact-clone curve.

`starship` losing a quarter of its composite is the clearest signal. That is not a
repository with 25 points of redundancy in it.

## The question that decides this

The triage that produced #102 enumerated 236 near-duplicate pairs across the
measured repos and **read none of them**. Everything measured so far is how much
the score *would* move, not whether it *should*.

Rust is both the worst-hit language here (median −3.70) and the language where
earlier duplicate-precision work measured the lowest precision, driven by trait
idioms — `fmt`, `new`, `visit_*` — that are structurally near-identical by
language design rather than by copy-paste.

If that is what is being penalised, this change does not correct an under-penalty.
It amplifies false positives into a large, silent score regression for every Rust
user.

## Recommendation

**Do not merge as written.** In order:

1. Label a sample of 0.7–0.95 similarity pairs, weighted toward Rust, and establish
   detector precision. This is the load-bearing unknown and nothing downstream is
   decidable without it.
2. If precision holds, retune the scale for near-duplicate semantics so the median
   lands near the v0.20.0 baseline rather than −2.1, and re-run the corpus.
3. If precision is poor, fix the detector before making its output louder.

The mechanism fix in this branch is sound and worth keeping. It should not ship on
the exact-clone curve.

---

src/drift/index.ts
```diff
@@ -323,6 +323,10 @@ export function driftFindingToFinding(d: DriftFinding): Finding {
     // they have no real dominance ratio, so we omit driftSignal and let the engine
     // size-normalize them through its count-based density branch instead of reading
     // a fabricated consistencyScore as a deviation rate.
+    // A count-based DUPLICATE detector still carries volume: `dupGroupSize`
+    // routes it to the engine's duplicate-fraction branch instead of the
+    // generic count branch, whose divisor is finding count (issue #102).
+    ...(d.dupGroupSize !== undefined ? { dupGroupSize: d.dupGroupSize } : {}),
```
BrandonGarate177, line 329: Anchoring here because this line is the cause. It's the first time a drift `Finding` carries
`dupGroupSize`, which makes an older unguarded line in the engine reachable.

Your point 2 says the dup-fraction branch "replaces `groupImportance` with a flat 1.0." Half
of that is by design, I think. On the composite path `groupDeviation` multiplies each group by
`findingMaxWeight` before dividing, so importance comes back inside the deviation. It moves
from a detector-level MAX to a per-group MAX. Finer, not weaker.

The bug is the path where routing *doesn't* happen. `driftFindingToFinding` has a second
caller at `index.ts:217` → `computeDriftScores` → `categoryHealth(catFindings, true, klocCount)`,
three args, so `functionCount` is 0. The dup branch is guarded off and bakes nothing in.
`detectorDamage` (`engine.ts:543`) isn't guarded and fires anyway:

```ts
const isDupDetector = findings.some((f) => (f.dupGroupSize ?? 0) > 1);
const importance = isDupDetector ? 1.0 : groupImportance(findings);
```

Its comment says importance is skipped "so this is not double-counted." On that path nothing
got counted in the first place.

Same fixture, three near-duplicate handlers, only the directory changes. Near-dups not exact,
so `codedna-fingerprint` stays out and redundancy has exactly one detector in all four cells:

| | redundancy (composite) | `semantic_duplication` bar |
|---|---|---|
| as submitted, `examples/` | 11.8/20 | **10.7/14** |
| as submitted, `src/` | 8.5/20 | **10.7/14** |
| with guard, `examples/` | 11.8/20 | 12.9/14 |
| with guard, `src/` | 8.5/20 | 10.7/14 |

Left column is your point 2, importance alive and worth 3.3 points. The guard doesn't touch
it. Right column is the bug: the bar reads 10.7 whether the dups are in `examples/` or in
shipped `src/`.

The engine line predates this PR, so fair to ask if it's a separate issue. I don't think so,
because the guard restores the old value rather than picking a new one:

```
base (7eb89b4)     12.9/14
branch (7dc228f)   10.7/14
branch + guard     12.9/14
```

Nothing could reach it before. `codedna-fingerprint` was the only detector carrying
`dupGroupSize` and it never flows through `computeDriftScores`.

Composite doesn't move, which is why the repo table wouldn't show it. CSV, DOCX and the
uploaded `driftScores` do.

Fix is one line in `engine.ts:543`, and `functionCount` is already a param there. Not a
suggestion block because that line isn't in this diff:

```ts
const isDupDetector = functionCount > 0 && findings.some((f) => (f.dupGroupSize ?? 0) > 1);
```

src/drift/semantic-duplication.ts
```diff
@@ -160,6 +160,42 @@ export const semanticDuplication: DriftDetector = {
...
+    for (const node of parent.keys()) {
+      const root = find(node);
+      const list = clusterMembers.get(root);
+      if (list) list.push(node);
+      else clusterMembers.set(root, [node]);
+    }
+    const redundantByDir = new Map<string, number>();
+    for (const members of clusterMembers.values()) {
+      if (members.length < 2) continue;
+      // fnKey is `${file}:${name}`; sort by code unit, never localeCompare.
```
BrandonGarate177, line 191: Comment has the key format wrong. `fnKey` is `` `${f.file}::${f.name}::${f.line}` `` (line
132), so `lastIndexOf(":")` gives you `src/a.ts::foo:`, not a path. Works only because
`directoryOf` reads the last `/` and ignores the rest.

Splitting on the first `::` says what it means. Ran it against the suite (8 passing) and
against an independent cluster-graph implementation over 16 topologies, including clusters
spanning three dirs and root-level files. Same attribution every time.

```suggestion
      // fnKey is `${file}::${name}::${line}`; sort by code unit, never localeCompare.
      const sorted = [...members].sort((a, b) => (a < b ? -1 : a > b ? 1 : 0));
      for (const member of sorted.slice(1)) {
        const dir = directoryOf(member.slice(0, member.indexOf("::")));
```

src/drift/semantic-duplication.ts
```diff
@@ -160,6 +160,42 @@ export const semanticDuplication: DriftDetector = {
...
+      const sorted = [...members].sort((a, b) => (a < b ? -1 : a > b ? 1 : 0));
```
BrandonGarate177, line 192: The comment promises determinism, "same input, same attribution, which the repo requires of
anything reaching a score." Deleting this sort passes all 8 tests:

```
unmutated                  8 passed
member count               3 failed | 5 passed    caught
pair count                 3 failed | 5 passed    caught
undercount 7x              2 failed | 6 passed    caught
step function              2 failed | 6 passed    caught
remove deterministic sort  8 passed               SURVIVES
```

Four of five caught is better than my first attempt at these tests managed, so this is the
only gap. One cross-dir cluster fed in two file orderings would pin it. Given
`discovery.ts:74-79` and the `localeCompare` history, probably the last property in the diff
you'd want unpinned.

CHANGELOG.md
```diff
@@ -4,6 +4,33 @@ All notable changes to `@vibedrift/cli` are documented here. The format
...
+  redundant *copies* rather than pairs, because a cluster of m mutually-duplicate
+  functions is m(m-1)/2 pairs but only m-1 redundant copies; and a cluster
+  spanning two directories is counted once rather than in both, by attributing
+  each copy to the directory that actually holds it.
+
+### Scoring
+
+**Composite scores move in this release and baselines rebuild.** `SCORING_VERSION`
+advances to v16. Repos with several near-duplicate functions in one directory
+score lower, which is the defect being corrected. Repos whose duplicate
```
BrandonGarate177, line 30: A single pair still crosses from the count branch into the dup-fraction branch, and the scales
differ (`COUNT_RATE_SCALE = 0.5` vs `DUP_FRACTION_SCALE = 0.15`), so it moves with zero volume
change. One pair, one directory, nothing else in the repo:

| | base | branch |
|---|---|---|
| 2 files, 1 pair | 74.8, redundancy 14.0 | 69.3, redundancy 12.0 |

Your corpus agrees, 192 of 212 moved.

```suggestion
score lower, which is the defect being corrected. Repos with a single pair per
directory move slightly, because a single pair still crosses from the count
branch to the duplicate-fraction branch and the two have different scales. Repos
with no near-duplicates are unchanged.
```

src/scoring/engine.ts
```diff
@@ -107,12 +107,23 @@ import {
...
+ *   v16: duplicate VOLUME reaches the score. `drift-semantic_duplication` was
+ *        emitted as count-based but carried no `dupGroupSize`, so the engine's
+ *        duplicate-fraction branch was unreachable for it and it fell through to
+ *        the generic count branch, whose divisor is the number of FINDINGS.
+ *        Findings are rolled up per directory, so one near-duplicate pair and
+ *        twenty in the same directory scored identically. The detector now
+ *        carries redundant copies per directory, attributing each cluster's
+ *        copies to the directory that holds them so a cluster spanning two
+ *        directories is not counted twice. Repos with several near-duplicates in
+ *        one directory score lower; repos whose duplicate directories hold a
```
BrandonGarate177, line 119: Same sentence, second copy. Worth fixing here too since this block is what the next person
reads to find out what v16 did.

```suggestion
 *        one directory score lower; a single pair per directory moves slightly,
 *        since it still crosses from the count branch to the duplicate-fraction
 *        branch and the two scales differ. Repos with no near-duplicates are
 *        unchanged. Reported in #102.
```

BrandonGarate177 (commented):
You're right `isNonShippablePath` is properly segment-anchored. The real mechanism is `allNonShippable()` getting `SourceFile.path`, which is absolute. Byte-identical content, only the ancestor directory differs:

```
.../fixtures/repo/src/f1.ts    codedna-fingerprint: 0 findings
.../plainname/repo/src/f1.ts   codedna-fingerprint: 1 finding
```

Worth its own issue. Anyone whose checkout lives under a `tests/`, `examples/`, `fixtures/`, `spec/`, `mocks/` or `generated/` segment silently loses exact-duplicate detection.

**Your corpus reproduces.** Cloned your worst movers, scanned each twice:

| repo | you | me |
|---|---|---|
| starship | 96.2 → 71.3 | 96.2 → 71.7 |
| underscore | 54.5 → 35.0 | 54.5 → 35.0 |
| htmx | 76.4 → 58.7 | 76.4 → 58.7 |
| moment | 68.2 → 52.1 | 68.2 → 52.1 |
| dayjs | 80.6 → 66.1 | 80.6 → 66.1 |

Four of five byte-identical. starship's 0.4 is probably a different clone commit.

### The 236 pairs nobody read

I read them, then measured the whole flagged population instead of a sample:

| repo | pairs | inline Rust `#[cfg(test)]` | same name, ≥.99 | genuinely near | hits a de-weighted path |
|---|---|---|---|---|---|
| starship | 10,383 | **93.0%** | 0.8% | 6.2% | **0.0%** |
| underscore | 852 | 0% | **89.3%** | 10.7% | **0.0%** |
| htmx | 501 | 0% | **100.0%** | 0.0% | **0.0%** |
| moment | 1,062 | 0% | **78.9%** | 11.3% | **0.0%** |
| dayjs | 39 | 0% | **74.4%** | 0.0% | **0.0%** |

starship is a test table. 1,452 of its 2,103 indexed functions are inline unit tests, and 93% of pairs have both ends inside a `#[cfg(test)]` module. Cross-checked content-side (both bodies contain an assertion) at 92.9%, so it's not an artifact of how I detect tests. 78.9% of its redundant copies come from those.

htmx ships three byte-identical copies of a 5,342-line file (`src/htmx.js`,
`www/static/src/htmx.js`, `www/themes/htmx-theme/static/js/htmx.js`, all md5 `143c79df…`). That's 100% of its pairs. underscore and moment are committed bundles. dayjs is `src/locale/*.js`, same declaration with different strings.

**The last column is the real finding.** htmx's duplicates are byte-identical, so "false
positive" is the wrong frame. The problem is that every population driving the new signal is something `computeFileImportanceWeight` exists to discount and structurally can't see, because it's path-based:

- Rust puts tests in `src/modules/php.rs`, so `TEST_PATH_RE` never sees a `tests/` segment.  (`isAnalyzableSource` drops anything containing `test` anyway, so `WEIGHT_TEST` is close to unreachable on this branch.)
- `www/static/src/htmx.js` and `underscore-umd.js` are build output with no `generated/` segment and no `.gen.` or `.min.` suffix. moment's sit under a `min/` directory, which that regex also misses.
- Locale data isn't test, example, or generated. It's a shape that repeats on purpose.

So the bigger lever looks like content-based de-weighting, not the scale. Inline-test
detection alone pulls 78.9% of starship's redundant copies out. A same-content-different-path vendor check takes basically all of htmx's. Retuning `DUP_FRACTION_SCALE` can't separate those from real drift, it can only make everything quieter.

Also worth saying: none of this is new. The detector already flagged all 10,383 starship pairs on v0.20.0. The count branch just couldn't see volume, so starship sat at 20.0/20 on redundancy while carrying them. This PR doesn't create the population, it prices it.

### Calibration

Both sides, same machine:

```
            0%     10%    25%    50%    75%    90%     0→10
base       86.7   87.4   86.6   80.9   73.3   71.5     +0.7   (red, pre-existing)
branch     61.9   71.6   71.0   65.5   57.6   56.2     +9.7   (red)
```

Red either way so it isn't introduced here, but the clean 0% drops 24.8 and the inversion grows ~14x. Same mechanism as the corpus: the baseline is 31 templated CRUD files where `UserRepository` and `OrderRepository` differ by a noun, and the injectors make files diverge, which breaks clusters, which lowers duplication damage. Ruled out the obvious alternative, `codedna-fingerprint` fires zero times on that corpus.

### The aggregation is correct

Since it looks like the risky part: I rebuilt the cluster graph with independent aggregation code and compared `Σ(dupGroupSize - 1)` plus the per-directory split across 16 topologies, including clusters spanning three dirs, a cluster whose canonical member is alone in its directory, and root-level files. Exact match on all of them. Build, 2,728 tests, lint and typecheck all clean, and your three tests bind (all fail with `src/` reverted to base).

### Summary

Mechanism's right, aggregation's exact, and I agree with your call not to ship it on the exact-clone curve. The guard and the two copies of the "single pair" sentence are worth doing either way. Happy to open those as a small branch.

Found four pre-existing things while measuring (`dedupePairs` undercount, lexicographic canonical bias, `findingMaxWeight` inflating the weighted copy count, and the phantom-scaffolding numbers not reproducing here). None of it belongs to this PR.

PR #109, opened 2026-08-30
feat(plugin): ship VibeDrift as a Claude Code plugin with /vibedrift:setup

anishekkamal:
## What

VibeDrift becomes an installable Claude Code plugin. The idea: make VibeDrift far more accessible — no multi-step CLI + MCP wiring, just a simple, fast install that gets anyone from zero to a working setup in under 30 seconds, without ever leaving the Claude Code terminal:

```
/plugin marketplace add VibeDrift/VibeDrift
/plugin install vibedrift@vibedrift
/vibedrift:setup   (once per repo)
```

## Changes

- **`.claude-plugin/plugin.json`** — bundles the MCP server (`npx -y @vibedrift/cli mcp`) so one install auto-wires all 8 MCP tools (no `claude mcp add` step), plus two skills: the existing in-loop drift-check skill and the new setup skill.
- **`.claude-plugin/marketplace.json`** — self-marketplace, so the two-command install above works today with no external listing. Community-marketplace submission (unlocks plain `claude plugins install vibedrift`) is a follow-up.
- **`skills/setup/SKILL.md`** — `/vibedrift:setup`, run once per repo, idempotent: surveys existing state (config, ignore file, managed CLAUDE.md block, installed hooks, duplicate user-level MCP server), configures excludes via the MCP `init` tool (detect → confirm → apply), runs the baseline scan + injects the managed context block into CLAUDE.md, and offers Drift Sessions — consent stays on the `enable` tool's native Allow/Deny prompt, never default-on.
- **`scripts/sync-plugin-version.mjs`** — keeps `plugin.json` version equal to `package.json`; runs in `release.sh` (same release commit/tag) and as a `--check` step in CI.
- **Docs** — README "Claude Code: 30-second setup" section, CHANGELOG notes, handbook chapter 09 (plugin channel), AGENTS.md MCP tool count corrected (6 → 8).

## Verification

- `claude plugin validate .` passes
- `node scripts/sync-plugin-version.mjs --check` passes (and correctly fails on a deliberately mismatched copy)
- `npm run lint` clean; `npm run handbook` rebuilt cleanly
- Reviewed: all doc/skill claims cross-checked against `src/cli`, `src/mcp`, `src/output/inject-context.ts` (flag names, tool schemas, gating, block markers, install commands consistent across README/CHANGELOG/handbook/skill)


BrandonGarate177 (commented):
Ran the branch. `plugin validate` passes, version sync works in all three modes, handbook rebuild is clean, and the `init`/`enable` contracts in the setup skill all match source.

One thing to fix: the skill opens with "Everything here is local." It isn't. I ran the step 3 command against a local capture server. Signed out it posts the beacon; signed in it also posts 6.5KB to `/v1/scans/log` carrying `locations[].snippet`, the literal source line. On a repo with a hardcoded secret, that uploads the secret. Disabling telemetry stops the beacon, not the upload. Captures and a suggested replacement are inline.

Not new behavior, not caused by this PR, but the skill now says the opposite exactly where the user is deciding what to consent to.

skills/setup/SKILL.md
```diff
@@ -0,0 +1,124 @@
...
+description: "One-time VibeDrift setup for this repo: configure excludes, build the drift baseline, write repo context into CLAUDE.md, and optionally enable Drift Sessions. Run once per repo after installing the vibedrift plugin."
+disable-model-invocation: true
+---
+
+# VibeDrift setup
+
+Run once per repo. Four steps: excludes, baseline + context, Drift Sessions,
+wrap-up. Everything here is local; the only network calls are the account check
+behind `--write-context` and, if the user enables Drift Sessions, the entitlement
+refresh.
```
BrandonGarate177: "Everything here is local" isn't true, and this is the paragraph the agent reads right before running step 3. That command carries neither `--local-only` nor a telemetry opt out.

Ran it against a local capture server (fake token, sandboxed `VIBEDRIFT_HOME`, nothing near production). Fixture holds AWS's documented example key on line 1.

Signed out:
```
POST /v1/beacon/scan     authed=false   192 bytes
```

Signed in, which step 3 tells the user to do:
```
GET  /account/credits    authed=true
POST /v1/beacon/scan     authed=false   192 bytes
POST /v1/scans/log       authed=true   6503 bytes
```

That last body, verbatim:
```json
"snippet":"const AWS_ACCESS_KEY_ID = \"AKIAIOSFODNN7EXAMPLE\";"
```

`sanitizeResultForUpload` drops `rootDir`, `ast` and `treeSitterNode` but keeps every finding, and `FileLocation.snippet` (`src/core/types.ts:467`) is a real source line.

Controls, since the escape hatches aren't obvious: `--local-only` signed in gives zero requests. `VIBEDRIFT_TELEMETRY_DISABLED=1` stops the beacon but `/v1/scans/log` still fires with the snippet. The beacon itself is clean (11 fields, no paths, no code).

Wording is yours, I just want the list to exist and to not imply the telemetry opt out covers the upload.

```suggestion
Run once per repo. Four steps: excludes, baseline + context, Drift Sessions,
wrap-up. Almost all of it is local. The exceptions, so you can answer honestly
if the user asks:

- `npx -y @vibedrift/cli` fetches the package from the npm registry.
- The step 3 scan sends an anonymous beacon (language, file count, LOC, score;
  no paths, no code) unless the user opted out with `vibedrift telemetry
  disable` or `VIBEDRIFT_TELEMETRY_DISABLED=1`.
- If the user is logged in, that same scan also uploads its findings to
  `/v1/scans/log`, and findings keep `locations[].snippet`, which is a real
  line of source. The telemetry opt out does not stop this. Only
  `--local-only`, or staying logged out, does.
- Enabling Drift Sessions refreshes entitlement from the server.
```

anishekkamal: > "Everything here is local" isn't true, and this is the paragraph the agent reads right before running step 3.

Confirmed against source and fixed in ec1425f — applied your suggested wording verbatim. Verified the chain myself before taking it: the upload is gated only on the bearer token (`scan.ts:670`), `--local-only` is the only network kill switch (`scan.ts:1150`), and `sanitizeNode`'s skip list (rootDir/ast/treeSitterNode) lets `locations[].snippet` through on `findings` (`sanitize-result.ts:194`). The telemetry opt-out covering only the beacon is exactly right. Thanks for running it against a capture server — that's the kind of review that earns its keep.

skhan75 (approved):
Verified by execution: every capability claim in the setup skill checked against the shipped code and all matched exactly — the init tool schema, the enable consent contract (confirm/decline/needs_confirmation, native Allow/Deny via requiresUserInteraction), the CLAUDE.md managed-block markers, the --write-context account gate string, both telemetry opt-outs, the #vibedrift-hook marker, --sync on, and the 8-tool count in the AGENTS.md correction. `claude plugin validate .` passes on both manifests, the sync script's 10 tests pass, a deliberate version mismatch fails --check with exit 1, the full suite is green (2735), and the committed handbook HTML matches a fresh rebuild except the embedded build date. The privacy disclosures are accurate against source, including the signed-in snippet-upload caveat.

The skill's behavior rules are exactly right for a setup flow: survey first, never write unconfirmed, stop when the MCP tools are absent instead of hand-editing config, user-invocation only. Clean work — welcome aboard.

---
🤖 Review verified and posted by Claude (Claude Code) on behalf of @skhan75

PR #113, opened 2026-09-01
fix: resolve the 2026-08-31 whole-repo audit findings (108 fixes across 9 areas)

anishekkamal:
## What this is

A whole-repo multi-agent audit (2026-08-31) surfaced 111 verified findings (3 P0, 29 P1, 48 P2, 31 P3). This PR fixes them — one branch as agreed, one commit per area for reviewability. Every P0/P1 fix carries a regression test; most were verified to bind by inverting the fix and watching the test go red.

## Headline fixes

- **Scoring monotonicity (P0)** — `src/scoring/engine.ts`: empty surface-specific categories dropped out of the geometric mean, so the *first* security/intent drift finding **raised** the Vibe Drift Score (measured 65.0 -> 74.9). Now capped so a finding can never raise the composite; property test asserts `composite(F) <= composite(F \ {f})`. `SCORING_VERSION` v15 -> v16.
- **CSV formula injection (P0)** — `src/output/csv.ts` + the HTML report's Export CSV helper: leading `=`/`+`/`-`/`@` now neutralized (CWE-1236).
- **Auth false-bless (P0)** — `src/drift/security-ast-python.ts`: a bare `raise Forbidden("maintenance")` no longer marks a whole receiver scope authenticated.
- **Upload privacy (P1)** — `src/ml-client/sanitize-result.ts`: complete function bodies (`rawBody`/`declarationCode`) were uploaded on every signed-in scan despite the module's "no file contents" contract; now stripped. (Note: the original audit finding blamed the ~120-char evidence `snippet`; investigation showed that is deliberate and test-pinned — the real leak was the function bodies.)
- **Report server LAN exposure (P1)** — binds `127.0.0.1` instead of `0.0.0.0`.
- Taint-analysis regex anchoring (a `print()` no longer silences SQL-injection flows; `RE.exec()` no longer flagged as command injection; interprocedural pass un-broken on TS/Go/Rust), WASM memory disposal, atomic session sidecar writes, watch liveness probing, evidence-ranked drift classification, hint-seeding fails closed, and ~90 more — see the per-commit messages.

## Reviewer decision point: calibration tradeoff

`npm run calibrate:monotonic` before/after:

| gate | main | this branch |
| --- | --- | --- |
| monotonicity (0% -> 10% must not raise score) | FAIL (86.7 -> 87.4) | PASS |
| responsiveness (each 25% step drops >= 3pt) | PASS | FAIL (25->50 drops 1.4) |

The responsiveness dip is largely intentional: at 50% injection no dominant pattern exists, and the new dominance gates follow AGENTS.md's own rule that "no norm exists" is not drift — the harness's expectation encodes the old ungated behavior. Options: accept and recalibrate the harness thresholds, or soften the dominance gate. Flagging rather than silently tuning.

`npm run calibrate` (precision/recall): all scored rows at 1.00/1.00, security-floor precision gate passes.

## Verification

- `npm run typecheck` clean, `npm run lint` clean, `npm run build` clean.
- Full vitest: **2969 passed / 46 failed — the 46 are byte-identical to `main`** (verified on a clean `main` worktree): Windows-environment issues (POSIX chmod/`/tmp` assumptions, `spawnSync` of `tsx` without shell) in 5 files untouched by this PR. Zero regressions from this diff.
- Two pre-existing tests were deliberately inverted because they pinned the bugs themselves (`bare raise PermissionDenied -> reject`, and the legacy regex-window over-bless); each carries a comment explaining why.
- One public-shape change: `DriftFinding` gains optional `allDominantFiles` (pivot detection needs the real dominant population, not the 3-file display sample).

https://claude.ai/code/session_013PuF7TRjbQrZwVwJoTbaKX

## Real-world score impact (reviewer-measured, 9 repos, base = pre-PR main)

| repo | base | this branch | change |
| --- | --- | --- | --- |
| moment | 68.2 | 56.6 | -11.6 |
| date-fns | 67.0 | 56.7 | -10.3 |
| zod | 75.7 | 67.4 | -8.3 |
| htmx | 76.4 | 68.4 | -8.0 |
| eslint | 85.2 | 79.4 | -5.8 |
| VibeDrift | 86.2 | 81.5 | -4.7 |
| starship | 71.7 | 97.2 | +25.5 |
| dayjs | 80.6 | 80.6 | 0.0 |
| underscore | 54.5 | 54.5 | 0.0 |

Nearly every score moves on upgrade. The direction is down or unchanged except where a previously over-penalised repo recovers. The starship jump does not reproduce on Windows (redundancy is 20/20 at the merge-base there too); see the open thread on `engine.ts`. Existing stored scores are kept and the one-time "scoring refined" notice shows after upgrade.

## Fix round (2026-09-01)

Rebased on main (v0.20.1). `SCORING_VERSION` v18, `imports` analyzer v4, `security` analyzer v4, `BASELINE_VERSION` 7. One fix commit per area on top of the area commit, so the stacked split is a cherry-pick per area:

- scoring: line-less ML duplicates merge with their Code DNA twin again; mixed signal/signal-less detector groups are monotone (max of dominance and count parts, sample-size discount folded in); randomized property test.
- drift: raw-SQL classification anchored to string literals; logging hints scoped to the languages that can honour them; convention-oscillation units fixed; BASELINE_VERSION 7.
- analyzers: Go unchecked-err reworked (no new flag conditions over the original heuristic); analyzer crashes print a warning and mark the result `degradedAnalyzers`; async body scan stops at overload signatures; security v4.
- session: resolved findings are tombstoned so a resolve survives the read-merge-write.
- ml-client: the scan beacon can never reject; the inline report-open beacon honours the telemetry opt-out; root-dir guard is MCP-only.
- output: `--format json` no longer embeds the raw CodeDnaResult.
- codedna: single-pass string/comment tokenizer; alias-receiver dotted sinks; LSH oversized buckets emit star+chain pairs; route-auth tree freed.
- docs: CHANGELOG `Unreleased`, handbook ch. 07/08.

Validation: typecheck and lint clean; 3049 tests pass, 46 failures are the known Windows-only set, unchanged from before.


src/scoring/engine.ts
```diff
@@ -107,12 +107,30 @@ import {
...
+ *        damage; it is now the group MAX, matching how severity and importance
+ *        already aggregate. Repos move only where one of those inversions was
+ *        active; the direction is always down or unchanged.
  *
  * A change here is absorbed silently for users: stored scores are re-aligned
  * where possible and a one-time release-notes notice is shown (see
  * src/core/scoring-notice.ts). Users never see this string.
  */
-export const SCORING_VERSION = "v15";
+export const SCORING_VERSION = "v16";
```
skhan75: this needs to be bumped to v18 as now there are changes merged to main, after rebasing your changes to latest main

src/scoring/engine.ts
```diff
@@ -456,6 +483,21 @@ function groupImportance(findings: Finding[]): number {
...
+ * deviation branch in `groupDeviation` AND the importance suppression in
+ * `detectorDamage` (the fraction branch bakes per-group file importance in via
+ * `findingMaxWeight`, so applying `groupImportance` on top would double-count).
+ * The fraction needs a real denominator, so a group with `dupGroupSize` but no
+ * known function count falls back to the per-KLOC density branch AND keeps
+ * detector-level importance — previously the two tests disagreed on exactly that
+ * case, dropping importance from a group that was not using the dup branch.
+ */
+function usesDupFraction(findings: Finding[], functionCount: number): boolean {
+  return functionCount > 0 && findings.some((f) => (f.dupGroupSize ?? 0) > 1);
```
skhan75, line 534: I measured this branch on 9 real repos and found something the PR body doesn't mention: on starship, the Redundancy category went from 11.1/20 to a perfect 20/20, while the report still shows error-severity findings saying "10,007 pairs of semantically duplicate functions". The same finding's damage went from 7.11 on main to 0.01 here.

I actually think the direction is right (we measured separately that ~90% of Rust "duplicate pairs" are false positives, so starship was being punished unfairly), but a perfect score displayed next to error findings from the same category is confusing to users.

Can you explain which change collapsed this damage, and whether it was intended? If it was intended, the report should probably stop showing those findings as errors.

---

src/analyzers/imports.ts
```diff
@@ -66,7 +66,7 @@ export const importsAnalyzer: Analyzer = {
   requiresAST: false,
   applicableLanguages: ["javascript", "typescript"],
   // Bumped when detection changes — invalidates the S1 findings cache.
-  version: 2,
+  version: 3,
```
skhan75: This needs to become **version: 4** after rebase. Main already has `version: 3` from #108 (a different change, itemCount). If both changes ship as v3, anyone who scanned with #108's build has a warm cache that will silently serve results without your ESM-stripping fix, and it won't reproduce on a clean machine. Same class of collision as the scoring version.

---

src/core/baseline.ts
```diff
@@ -293,6 +293,10 @@ export async function buildBaseline(rootDir: string): Promise<RepoDriftBaseline>
   const { ctx } = await buildAnalysisContext(rootDir);
   await parseFiles(ctx.files);
   const { driftFindings } = runDriftDetection(ctx);
+  // assembleBaseline below reads ctx.files' content/paths only (content
+  // hashing, extractAllFunctions — neither touches the AST) — drift
+  // detection above was the last tree reader, so it's safe to free here.
+  disposeTrees(ctx.files);
```
skhan75, line 305: Separate from this line: I don't see a `BASELINE_VERSION` bump anywhere in the branch (it's still 5; main is at 6 now). This PR changes vote and classification logic in `src/drift/utils.ts` and adds `allDominantFiles` to `DriftFinding`. If any of that changes what cached baselines store or how dominance is computed, the version must bump (to 7 after rebase), because a stale baseline never rebuilds on content changes alone — a version mismatch is the only trigger.

If you believe no stored shape changed, please say so explicitly in the PR body so we're choosing that on purpose.

---

src/analyzers/security.ts
```diff
@@ -238,7 +238,13 @@ const SECURITY_PATTERNS: SecurityPattern[] = [
...
     id: "python-yaml-unsafe",
     name: "Unsafe YAML load",
-    pattern: /yaml\.load\s*\([^)]*(?!\bLoader\b)/g,
+    // No in-pattern negative lookahead for `Loader=` here: `[^)]*(?!\bLoader\b)`
+    // is a no-op — `[^)]*` is greedy but backtracks freely, so it can always
+    // shrink by one character to satisfy the lookahead regardless of whether
+    // "Loader" appears later in the call. The `negativeFilter` below (tested
+    // against the full matched line, not constrained by backtracking) is
+    // what actually excludes `yaml.load(f, Loader=SafeLoader)` etc.
+    pattern: /yaml\.load\s*\(/g,
```
skhan75, line 247: Nice catch on the lookahead being a no-op. But this analyzer's `version` stayed at 3 while every other touched analyzer got bumped. Even if the finding set is identical because `negativeFilter` was already doing the work, the matched text can differ, and a version bump is one integer. Please bump to 4 so warm caches can't serve mixed results.

---

skhan75 (commented):
This is a strong PR and the verification section is honest. I reproduced the calibration tradeoff exactly as described, confirmed the three headline fixes bind by inverting them (scoring cap: 4 tests go red, CSV injection: 12, Python bless: 9), and confirmed the 46 test failures are Windows-only (everything passes on macOS and Linux, 3027/3027 here).

Two things before this can merge, on top of the inline comments:

**1. The real-world score impact is much bigger than the PR body suggests, and it needs to be in the record.** I scanned 9 real repos on this branch versus its base:

| repo | base | this branch | change |
| --- | --- | --- | --- |
| moment | 68.2 | 56.6 | -11.6 |
| date-fns | 67.0 | 56.7 | -10.3 |
| zod | 75.7 | 67.4 | -8.3 |
| htmx | 76.4 | 68.4 | -8.0 |
| eslint | 85.2 | 79.4 | -5.8 |
| VibeDrift | 86.2 | 81.5 | -4.7 |
| starship | 71.7 | 97.2 | **+25.5** |
| dayjs | 80.6 | 80.6 | 0.0 |
| underscore | 54.5 | 54.5 | 0.0 |

Nearly every user's score will visibly move on upgrade. That may be fine (the fix is correct, and the old scale was known to be compressed), but a shift this size is a product decision, not just a bug fix, and the starship jump needs the explanation I asked for on the engine change.

**2. The calibration decision point:** I'll rule on the monotonicity-versus-responsiveness tradeoff separately. My current lean is to accept it — monotonicity is a correctness property, and the responsiveness thresholds encode the old ungated behavior — but hold off on recalibrating the harness until I confirm.

A deeper multi-expert pass over each commit area is still in flight; anything it confirms will follow as additional comments.

---

src/scoring/dedup.ts
```diff
@@ -77,21 +85,33 @@ export function deduplicateFindingsAcrossLayers(findings: Finding[]): Finding[]
...
+ * way regardless of the order a layer listed it in.
  */
 function makeFilePairKey(f: Finding): string {
-  // Use ALL unique files (sorted + deduped) so 3+-file clusters merge
-  // correctly when reported by different layers. The old approach used
-  // only the first two files, which meant {A,B,C} and {A,C} could
-  // produce different keys and inflate the duplicate count.
   const files = [...new Set(
     f.locations.map((l) => l.file).filter(Boolean),
   )].sort();
```
skhan75: **P1 — the new dedup key breaks deep-scan merging.** The key is now `${file}#${line ?? ""}`, but ml-duplicate findings are built with no line field (`src/ml-client/confidence.ts:26-28`), while codedna findings always have one. So the ML copy and the Code DNA copy of the same clone no longer merge: deep scans show the clone twice, the "[confirmed by ...]" annotation is gone, and the clone's damage counts double in Redundancy. We reproduced this with production-shaped findings. The new test doesn't catch it because its fixture gives the ML finding a line number, which the real constructor never does.

---

src/scoring/engine.ts
```diff
@@ -456,6 +483,21 @@ function groupImportance(findings: Finding[]): number {
...
+}
+
```
skhan75, line 536: **P1 — the monotonicity guarantee has an untested hole.** The proof depends on every finding group under one analyzerId being all-driftSignal or all-not. A mixed group flips `findings.every((f) => f.driftSignal)` and lands in the count branch, and we measured a +31.3-point score inversion from adding one signal-less finding to a dominance group. No production detector emits a mixed group today, so this is latent, but the property test should pin the invariant (or the branch should handle mixed groups) so the next detector author can't silently break the guarantee this PR exists to establish.

---

src/drift/architectural-contradiction.ts
```diff
@@ -45,16 +46,37 @@ function detectDataAccess(file: DriftFile): { pattern: DataAccessPattern; eviden
...
+  // most-distinctive-deviation-first, matching DATA_ACCESS_PRIORITY below.
+
+  // Raw SQL. The clause keywords are separated by the column/table list, so
+  // each statement form is matched as a keyword PAIR within a bounded window
+  // rather than as adjacent tokens. The old
+  // `(?:SELECT|INSERT|UPDATE|DELETE)\s+(?:FROM|INTO|SET|\*)\b` matched none of
+  // `SELECT * FROM users` (a `\b` after `*` cannot hold before a space),
+  // `SELECT id FROM users` (FROM had to follow SELECT immediately), or
+  // `UPDATE users SET x = 1`.
+  const sqlEvidence = extractEvidence(
```
skhan75: **P1 — the rewritten raw-SQL regex over-matches badly.** `\bSELECT\b[\s\S]{0,200}?\bFROM\b` (case-insensitive) matches import lines and prose comments. Executed on a 4-repo population: a majority of its file classifications are false, and this repo's own `src/output/html.ts` and `src/output/terminal.ts` classify as "raw SQL queries". The new push-order tie-break then makes raw_sql win those ties. This needs word-boundary-plus-context anchoring or an AST check before it ships.

---

skhan75 (commented):
Follow-up from the deep multi-expert pass (10 area reviewers, every finding adversarially re-executed by an independent verifier; 28 confirmed, 0 refuted). The 8 inline comments above are the confirmed P1s — each was reproduced by running the real code, not by reading it.

Confirmed P2s, grouped, briefly (details available on request):

**Required gating:**
- `BASELINE_VERSION` must bump (three independent verifiers confirmed: the tokenizer change alters persisted MinHash signature streams, and the drift + security vote logic changed). This upgrades my earlier "if" comment to a definite requirement.
- No CHANGELOG entry, and Developer Handbook ch. 07/08 are now stale (confidence mean→max, deleted deviation-weight override still documented) — the repo rule is same-PR updates.

**Real defects, lower blast radius:**
- The tokenizer string-before-comment reorder fixes URLs-in-strings but introduces the mirror bug: an apostrophe in a comment ("don't") pairs with the next quote and swallows real code.
- The bare taint anchor `(?<![\w$.])` un-detects real dotted sinks: `cp.exec(cmd)` and `window.fetch(url)` no longer match.
- `MAX_LSH_BUCKET=200` silently drops ALL candidate pairs for clusters of 200+ near-identical functions (the comment's recovery claim doesn't hold — every band shares the same oversized bucket).
- codedna-pattern's new driftSignal is defeated whenever the same scan emits a mixed-patterns finding (shared analyzerId, one signal-less finding flips the group's branch — same mechanism as the engine comment above).
- convention-oscillation now divides a symbol-count numerator by a file-count denominator; the rendered "X of Y" can exceed 100%.
- The error-handling async pattern matches body-less TS overloads/abstract signatures and scans 500 chars past the `;` into the next function.
- `--format json` now dumps the whole CodeDnaResult, embedding every extracted function's full source body — roughly doubles output size and puts complete source in a surface that never carried it.
- The WASM fix is incomplete for the long-lived MCP server: `route-auth-classify.ts` parses and never deletes its tree, so validate-change still leaks one tree per call (file untouched by this PR, so noting it here).
- The report-open beacon opt-out fix lands on a function with zero production callers; the real beacon is inline in the generated HTML and still ignores the opt-out.
- The new root-dir guard also gates the interactive `vibedrift init` wizard, which now refuses at the end of the walkthrough in a directory without the expected markers.

To be clear on the other side of the ledger: the panel also verified a long list of things as correct — the dedup copy contract survives, the silent-migration machinery is intact, monotonicity held in 4,400+ randomized removal trials on production-reachable shapes, the security-lane bless changes respect NEVER-FALSE-BLESS, and the two inverted tests were both judged justified. The core of this PR is good work. It needs a fix round for the items above plus the rebase (v18 / imports v4 / BASELINE_VERSION 7) before it can land.

---

src/drift/logging-consistency.ts
```diff
@@ -116,64 +117,101 @@ export const loggingConsistency: DriftDetector = {
...
+  // Files that use multiple families get classified by the most-used one.
+  // When an intent hint declares a logger (e.g. "use winston"), seed the
+  // dominance vote so the declaration carries weight even when the
+  // codebase is in transition. Vocabulary guard: a seeded vote bypasses the
+  // 60% dominance gate below and `seedDominanceVote` injects an unknown
+  // declared pattern as a ZERO-COUNT entry weighted ~1.95, so a hint outside
+  // this axis's families could win the vote outright — reporting a dominant no
+  // file uses, with consistencyScore 0. Only a key of FAMILY_NAMES may seed.
+  const counts = buildPatternDistribution(profiles);
+  const rawHint = pickIntentHint(ctx, "logging_consistency");
```
skhan75, line 203: **P1 — the language partition misapplies repo-wide hints.** A language-exclusive declaration like `go_slog` in CLAUDE.md now emits a false "team declared X but files use Y" divergence finding on every OTHER language's partition, with a "Migrate 0 file(s)" recommendation. Reproduced: 3 Go slog files + 3 TS console files + one go_slog hint produces the false finding on the TypeScript partition. The hint needs to be scoped to the language it belongs to.

---

src/analyzers/language-specific.ts
```diff
@@ -38,18 +40,25 @@ function detectGoUncheckedErrors(
...
-        let nextLine = "";
-        for (let j = i + 1; j < Math.min(i + 4, lines.length); j++) {
+        let checked = false;
+        let hasFollowingContent = false;
+        for (let j = i + 1; j < Math.min(i + 6, lines.length); j++) {
           const next = lines[j].trim();
-          if (next && !next.startsWith("//")) {
-            nextLine = next;
+          if (!next || next.startsWith("//")) continue;
+          hasFollowingContent = true;
```
skhan75, line 69: **P1 — the Go unchecked-err rewrite increases false flags 41%.** The commit message says this reduces false flags; measured on 505 real Go files it goes 447 → 632 flags (+41%), and a 15-item sample of the new flags was 15/15 false positives on the most idiomatic Go error shapes. This one should be reverted or reworked before merge.

---

src/core/run-analyzers.ts
```diff
@@ -40,7 +41,15 @@ export async function runAnalyzers(
...
-        findings = await analyzer.analyze(ctx);
+        // Isolate each analyzer: one throwing must not reject the whole
+        // Promise.all and kill every other analyzer's findings. Log and
+        // skip the failing one, keep going with the rest.
+        try {
+          findings = await analyzer.analyze(ctx);
+        } catch (err) {
+          debug("analyzers", `analyzer "${analyzer.id}" threw and was skipped:`, err);
+          return { findings: [], hit: false };
+        }
```
skhan75, line 68: **P1 — analyzer crashes are now completely silent for normal users.** The only surface is `debug()`, which is off unless `VIBEDRIFT_DEBUG` or `--verbose`. Executed: a throwing analyzer produces zero stderr, exit 0, and a score computed as if it ran clean (it's even tallied as an ordinary cacheMiss). Crash isolation is right, but the user must see something — a warning line and/or a `degraded` marker in the result. A silently-wrong score is the exact outcome this project's audit-first rule exists to prevent.

---

src/session/outcomes.ts
```diff
@@ -77,6 +78,42 @@ export function emptyOutcomeState(): OutcomeState {
...
+ *  finding costs a re-flag; a false resolve costs the user's trust"), so it
+ *  is not a new risk, just the existing bias applied here too. What a union
+ *  actually prevents is the real hazard: a flag one process just appended to
+ *  `open` silently vanishing because the other process's write landed last
+ *  with a state that never knew about it.
+ *
+ *  `hashes` unions each file's recent-body-hash list and re-caps it at
+ *  MAX_HASHES_PER_FILE, so revert detection never loses a hash either writer
+ *  recorded. */
+export function mergeOutcomeState(local: OutcomeState, onDisk: OutcomeState): OutcomeState {
```
skhan75, line 121: **P1 — resolves never persist.** `writeOutcomeState` always re-reads disk and merges, and the `open` merge is a pure union, so a finding you just resolved and filtered out is resurrected from disk on the very write meant to remove it. Reproduced deterministically with a single writer: flag DF-1, resolve it, write — the sidecar still contains DF-1. The merge needs a tombstone (or the resolved set must win over the disk copy).

---

src/telemetry/beacon.ts
```diff
@@ -113,13 +113,21 @@ export async function sendScanBeacon(
   try {
```
skhan75: **P1 — a malformed apiUrl now crashes every scan.** `resolveApiUrl()` newly throws on non-https or malformed URLs, and it's awaited at line 110, OUTSIDE the try that starts at line 113. `scan.ts` fires the beacon as `void sendScanBeacon(...)`, so the throw becomes an unhandled promise rejection and kills the process — for free, signed-out, local scans included, for anyone with a stale `VIBEDRIFT_API_URL` or config apiUrl. Move the await inside the try (and the beacon should never be able to crash a scan at all).

---

skhan75:
Stepping back from the individual findings: this PR is too large to review and land safely as one unit — 146 files across 9 independent areas, and the confirmed issues above are spread over 6 of them. A fix round on the monolith means re-reviewing everything again. Since you already structured it as one commit per area, please split it into stacked, independently testable PRs. Concretely:

**Land fast (low risk, panel-verified sound, minimal coupling):**
1. `7c9c655` output — CSV formula injection + renderer alignment. Drop the `--format json` full-CodeDnaResult dump from it (flagged above) and this can merge almost immediately.
2. `c17b106` security lane — the bless/flag fixes were verified against NEVER-FALSE-BLESS and both test inversions were judged justified. Ship it on its own.
3. `83fba02` cli/core — loopback bind + WASM disposal + watch honesty. Complete the WASM fix for the MCP server path (route-auth-classify) while you're in there.
4. `74277ef` session — atomic sidecar writes, with the resolve-resurrection fix (P1 above) folded in.
5. `ce40bfd` ml-client/privacy — body stripping + auth hardening, with the beacon-crash fix (P1 above) and the init-wizard guard scope-down.

**Middle (real fixes needing rework first):**
6. `d49757f` analyzers — rework or drop the Go unchecked-err change (measured +41% false flags), give crash isolation a user-visible surface, then land.
7. `899f801` codedna — fix the tokenizer apostrophe mirror-bug, the dotted-sink taint gaps, and the LSH bucket-cap hole first.
8. `3a23f00` drift classification — fix the raw-SQL regex and the logging hint partition (both P1 above), and this is the PR that carries the BASELINE_VERSION bump.

**Last, on its own:**
9. `60d2ab7` scoring — this one reshapes every user's score (measured -12 to +25 on real repos) and carries the calibration tradeoff, so it needs a maintainer product decision, its own SCORING_VERSION (v18 after rebase), the dedup-key fix, and the mixed-group invariant pinned. It should not be coupled to anything else.

Split the `83937b2` test/lint commit into whichever PRs its pieces belong to. Each split PR keeps its own regression tests (you already wrote them per area) and its own measurement section. Order matters mainly for 8 and 9 (version bumps rebase cleanly when they go last).

The work itself is genuinely good — this is about landing it safely, not about the quality. Happy to review the splits as they come.

---

skhan75 (changes requested):

src/scoring/engine.ts
anishekkamal: > this needs to be bumped to v18 as now there are changes merged to main, after rebasing your changes to latest main

Addressed: rebased onto main (v0.20.1, which carries v16 and v17); this branch's monotonicity repair is now `SCORING_VERSION = "v18"`, with its history entry appended after v17.

src/analyzers/imports.ts
anishekkamal: > This needs to become **version: 4** after rebase. Main already has `version: 3` from #108

Addressed: `importsAnalyzer.version` is 4 after the rebase (v3 = itemCount from #108, v4 = hasESM tested against comment-stripped content), and the pinned-version test asserts 4.

src/core/baseline.ts
anishekkamal: > BASELINE_VERSION must bump (three independent verifiers confirmed: the tokenizer change alters persisted MinHash signature streams, and the drift + security vote logic changed).

Addressed: `BASELINE_VERSION` is now `7` (the branch is rebased on main's 6). The doc comment gained a version-history line stating why: the codedna tokenizer change alters every persisted MinHash signature stream, the drift and security vote logic changed, and `DriftFinding` gained `allDominantFiles` so a v6 vote shape is missing fields.

src/analyzers/security.ts
anishekkamal: > this analyzer's version stayed at 3 while every other touched analyzer got bumped ... Please bump to 4 so warm caches can't serve mixed results.

Addressed: bumped `securityAnalyzer.version` from 3 to 4, with a comment line explaining why (python-yaml-unsafe pattern text changed; the finding set is expected to be identical since negativeFilter already filtered, but matched snippet text can differ, so warm caches must not serve a mix).

src/scoring/dedup.ts
anishekkamal: > the ML copy and the Code DNA copy of the same clone no longer merge: deep scans show the clone twice, the '[confirmed by ...]' annotation is gone, and the clone's damage counts double in Redundancy.

Addressed: good catch. `filterByConfidence` builds ML duplicates with file-only locations, so a `file#line` key could never match them. The grouping is now two-pass: findings whose locations all carry lines keep the site key (so N distinct clones over the same file pair still survive as N); a finding with any line-less location merges by file set into the first lined group over those files (first in sorted key order, so deterministic), and only forms its own file-set group when no lined group exists. The cross-layer merge test fixture now builds the ML finding exactly as the real constructor does (no lines) and asserts a single `ml-duplicate` survivor annotated with `[confirmed by ...]`; two new tests pin "attaches to exactly one of three lined groups, the other two survive" and "line-less findings over the same file set merge with each other when no lined group exists".

src/scoring/engine.ts
anishekkamal: > A mixed group flips `findings.every((f) => f.driftSignal)` and lands in the count branch, and we measured a +31.3-point score inversion from adding one signal-less finding to a dominance group.

Addressed differently: rather than only pinning the invariant, `groupDeviation` now handles mixed groups, since this is reachable today (`codedna-pattern` emits "Pattern drift" with `driftSignal` and "Mixed patterns" without, under one analyzerId; that also closes the P2 you listed for it). The function is split into a dominance part (over the signalled findings) and a count part (over the signal-less ones, keeping the `usesDupFraction` / per-function / per-KLOC semantics); a mixed group takes the max of the two, so adding a finding of either kind can only raise it. Pure groups are byte-identical to before.

Writing the randomized property test surfaced a second hole in the same mechanism: `groupSampleConfidence` discounted the whole mixed group by the dominance sample size, so removing the last signalled finding lifted the discount to 1.0 and raised the damage (measured 83.7 -> 68.6 from closing one finding). The discount is now folded into the dominance term for mixed groups; pure dominance groups are discounted exactly as before.

Tests added to `monotonicity.test.ts`: your repro shape (a `codedna-pattern` dominance group plus one "Mixed patterns" finding never raises the composite) and a seeded 200-trial property check over random mixed groups asserting that removing any single finding never lowers the composite. The property test fails against the max-only version and passes with the sample-discount fix.

src/drift/architectural-contradiction.ts
anishekkamal: > `\bSELECT\b[\s\S]{0,200}?\bFROM\b` (case-insensitive) matches import lines and prose comments ... this repo's own src/output/html.ts and src/output/terminal.ts classify as 'raw SQL queries'.

Addressed: the raw-SQL regex is now anchored to the start of a string literal. Each statement form must be preceded by `"`, `'`, or a backtick plus optional whitespace (so Python `"""..."""` and Go raw strings still qualify, including ones that open on a newline), and its clause target must be an identifier-ish token (`SELECT ... FROM <ident>`, `INSERT INTO <ident>`, `UPDATE <ident> SET`, `DELETE FROM <ident>`). Keyword case is all-upper or all-lower per statement rather than `i`, so a sentence-case UI string like `"Select an item from the list"` no longer matches while `db.query("select * from users")` still does. Confirmed the concrete false positive: the old regex bridged `"./select.js";\nimport ... from` across two import lines. New tests: import lines and prose comments/UI strings do not classify; `src/output/html.ts` and `src/output/terminal.ts` are read from disk and asserted not raw SQL via both `hasDataAccessPattern` and `classifyDataAccessLabel`; ten real SQL-in-string shapes (plain, multi-line template, Go raw, Python triple-quote, lowercase, `${table}`, UPDATE/INSERT/DELETE) still classify.

src/drift/logging-consistency.ts
anishekkamal: > A language-exclusive declaration like `go_slog` in CLAUDE.md now emits a false 'team declared X but files use Y' divergence finding on every OTHER language's partition, with a 'Migrate 0 file(s)' recommendation.

Addressed: added `FAMILY_LANGUAGES` beside `FAMILY_NAMES` (`python_logging` -> python, `go_slog` -> go, `debug_pkg`/`console` -> js/ts, `structured` -> any language since `logger.*` wrappers and Rust `tracing::` both land there). `voteWithinLanguage` now receives the partition's language and only accepts the hint when its pattern is a known family and that family is reachable from the partition's language. Regression test reproduces your shape (3 Go slog files + 3 TS console files + a `go_slog` hint at 0.9) and asserts zero findings.

src/analyzers/language-specific.ts
anishekkamal: > the Go unchecked-err rewrite increases false flags 41% ... a 15-item sample of the new flags was 15/15 false positives on the most idiomatic Go error shapes.

Addressed differently: reworked rather than reverted, so it adds no new flag conditions over the original heuristic. (1) An assignment line that itself contains `err != nil` / `err == nil` (the `if err := f(); err != nil {` / `for` / `switch` init forms) is treated as checked and skipped outright. (2) Otherwise, any following non-comment line in the ~5-line window that references `\berr\b` counts as handling it, the original rule just windowed, so `log.Fatal(err)`, `t.Fatal(err)`, `panic(err)`, `errors.Is(err, ...)`, `return wrap(v), err` are no longer flagged. The requirement that the handling line start with if/switch/return is gone; that was the source of the +41%. The comment now states that the only reduction in false flags vs. the old single-line lookahead comes from the window (the `n, err := f(); _ = n; if err != nil` shape). Tests cover each of those idioms (must not flag) plus a genuinely unchecked `err := f()` / `fmt.Println("done")` / `return nil` (must flag). One deliberate difference from the old code: a bare `return` without `err` on the next line is no longer an exemption, because `err := f()` followed by `return nil` does discard the error.

src/core/run-analyzers.ts
anishekkamal: > analyzer crashes are now completely silent for normal users ... the user must see something, a warning line and/or a degraded marker in the result.

Addressed: both. `runAnalyzers` now always writes one line to stderr on catch (`vibedrift: warning - analyzer "<id>" failed and was skipped; the score does not include its findings (<message>)`), keeps the stack behind `debug()`, returns a new `failed: string[]` on `RunAnalyzersResult`, and no longer tallies the failure as a cache miss. `scan.ts` threads that into the result as an optional `ScanResult.degradedAnalyzers` (set only when non-empty) and `renderJsonOutput` passes it through, so `--format json` / CI consumers see the marker. Tests assert `failed` is `["boom"]`, `cacheMisses` excludes it, and stderr received the warning.

src/session/outcomes.ts
anishekkamal: > resolves never persist. `writeOutcomeState` always re-reads disk and merges, and the `open` merge is a pure union, so a finding you just resolved and filtered out is resurrected from disk on the very write meant to remove it. [...] The merge needs a tombstone (or the resolved set must win over the disk copy).

Addressed: added tombstones. `OutcomeState` now carries `resolved: string[]` (finding ids, most recent last, capped at 200 the way `hashes` is capped). `mergeOutcomeState` builds `open` as (`onDisk.open` minus ids in `local.resolved`) union `local.open`, and `resolved` as (`onDisk.resolved` union `local.resolved`) minus anything in the merged `open`, so a re-flag clears its tombstone. `readOutcomeState` treats a missing `resolved` key (older sidecars) as empty. In `hook-entry.ts`, the resolve path pushes the resolved ids onto `outcomes.resolved` where they are filtered out of `open`, and the flag path drops the flag's id from `resolved` before pushing it open.

The tombstone is deliberately one-directional: an id in `local.open` survives even if `onDisk.resolved` names it, which keeps the module's existing "a stale open costs a re-flag; a false resolve costs trust" bias, and the doc comment now says so. Tests: your single-writer repro (write DF-1 open, read, resolve, write, read back: `open` is `["DF-2"]`, `resolved` is `["DF-1"]`), re-flag after resolve reopens and clears the tombstone, a legacy sidecar with no `resolved` key still reads, plus the pure-merge cases. Also drove the real hook end to end (flag -> fix -> re-flag): the sidecar goes `{open:[DF-1],resolved:[]}` -> `{open:[],resolved:[DF-1]}` -> `{open:[DF-2],resolved:[DF-1]}`.

src/telemetry/beacon.ts
anishekkamal: > resolveApiUrl() newly throws on non-https or malformed URLs, and it's awaited at line 110, OUTSIDE the try that starts at line 113 ... the throw becomes an unhandled promise rejection and kills the process

Addressed: the entire body of `sendScanBeacon` (the `isTelemetryEnabled()` config read, `resolveApiUrl()`, and the fetch) now runs inside one catch-all `try`, so the function can never reject regardless of what `VIBEDRIFT_API_URL` / config `apiUrl` contains. The timer-clearing `finally` is kept and the doc comment states the never-reject invariant. Regression tests: with `VIBEDRIFT_API_URL="http://evil.example"` and with `"not a url"`, `await sendScanBeacon(payload)` resolves and fetch is not called; a rejecting `readConfig` also resolves.

src/scoring/engine.ts
anishekkamal: > Can you explain which change collapsed this damage, and whether it was intended?

I re-ran starship on Windows at the merge-base (7eb89b4), at each main commit since (#107, #108, v0.20.1) and on this branch. Every one of them reports Redundancy 20/20 with the same three error findings (duplicates: 10,406 pairs; semantic-duplication: src/configs 394 pairs, src/modules 9,989 pairs; consistencyImpact 0.01 each), composite 96.2 -> 97.2. So on my machine the collapse predates this PR and I can't reproduce the 11.1 base you measured; most likely a platform difference (same class as the 46 Windows-only test failures).

Mechanism as far as I can trace it: those findings are count-based (semantic-duplication is `countBased: true`, `duplicates` emits one project-level finding with no `dupGroupSize`), so each finding contributes one item. Ten thousand pairs rolled into one finding read as 1/functionCount, so deviation rounds to zero. Nothing in this PR touches that branch of the engine; the only change that affects count groups is confidence mean -> max, which can only raise damage.

Making the volume register would mean carrying `itemCount` (distinct duplicated functions) on those findings through the #108 machinery. That re-penalises starship, where you measured ~90% of the Rust pairs as false, so it's a calibration call rather than a bug fix and I've left it out of this round. Could you share the per-detector damage from your base run so we can pin the platform delta?

anishekkamal:
Follow-up on the confirmed P2 list (the eight inline P1s are answered on their threads). Everything below is on the branch, rebased on main, one fix commit per area.

> `BASELINE_VERSION` must bump

Addressed: 7, with the reason in the version-history comment (persisted MinHash signatures and vote shapes changed).

> No CHANGELOG entry, and Developer Handbook ch. 07/08 are now stale

Addressed: `## Unreleased` in CHANGELOG.md (with the nine-repo table); ch. 08 documents confidence as the group max, the clean-credit ceiling and the surface-specific clamp; ch. 07 drops the `.vibedrift.json` override and `git_recency`, adds `no_comment`, and describes the single-pass tokenizer and the LSH bucket policy. Handbook rebuilt.

> The tokenizer string-before-comment reorder ... introduces the mirror bug: an apostrophe in a comment ("don't") pairs with the next quote

Addressed: neither sequential order is correct. Both tokenizers now recognise strings and comments in one left-to-right alternation shared between `minhash.ts` and the function extractor, so whichever construct starts first wins. Tests pin both directions (URL in a string, apostrophe in a comment, `//` inside a block comment).

> The bare taint anchor `(?<![\w$.])` un-detects real dotted sinks: `cp.exec(cmd)` and `window.fetch(url)` no longer match.

Addressed: the bare anchor stays (still excludes `RE.exec(`, `parseFunction(`, `reopen(`) and the dotted form is accepted when the receiver is a known alias (`child_process|childProcess|cp|proc|subprocess|shell|sh` for exec/execSync, `window|globalThis|self|global|node` for fetch). The generic `child_process` sink no longer double-reports `.exec(`.

> `MAX_LSH_BUCKET=200` silently drops ALL candidate pairs for clusters of 200+ near-identical functions

Addressed: correct, identical signatures collide in every band. An oversized bucket now emits star+chain pairs (~2m) instead of being skipped, which keeps the cluster connected for grouping at O(m). Test: 250 identical signatures, every index covered, 497 pairs.

> codedna-pattern's new driftSignal is defeated whenever the same scan emits a mixed-patterns finding

Addressed in the engine: a group that mixes signalled and signal-less findings scores as the max of its dominance and count parts (see the mixed-group thread).

> convention-oscillation now divides a symbol-count numerator by a file-count denominator

Addressed: `totalRelevantFiles` is a symbol count again, matching `dominantCount`.

> The error-handling async pattern matches body-less TS overloads/abstract signatures and scans 500 chars past the `;`

Addressed: the body scan stops at a depth-0 `;`. Test with overload signatures pins the exact finding count (was 10, now 4).

> `--format json` now dumps the whole CodeDnaResult, embedding every extracted function's full source body

Addressed: dropped from the JSON output. Drift findings and reimplementation candidates stay. Test asserts a function body never reaches the JSON.

> The WASM fix is incomplete for the long-lived MCP server: `route-auth-classify.ts` parses and never deletes its tree

Addressed: `classifyRouteAuth` deletes its tree in a `finally`; a spy on `Tree.prototype.delete` asserts exactly one delete per call.

> The report-open beacon opt-out fix lands on a function with zero production callers; the real beacon is inline in the generated HTML

Addressed: the dead `sendReportOpenBeacon` and its tests are gone; `scan.ts` passes `scanId` into the HTML renderer only when telemetry is enabled, so the inline beacon honours the opt-out.

> The new root-dir guard also gates the interactive `vibedrift init` wizard

Addressed: the guard moved from the shared init core to the MCP adapter, so `vibedrift init` can initialise a marker-less new project while the MCP `init` tool still refuses one.

Validation on this branch: typecheck and lint clean; 3049 tests pass, the 46 failures are the known Windows-only set (integration tests spawning the `.bin/tsx` shim, trial-recap, upload-follower), unchanged from before this round.

anishekkamal:
> please split it into stacked, independently testable PRs

Fix round first, on this branch: rebased on main (v18 / imports v4 / security v4 / BASELINE_VERSION 7), every inline P1 and the P2 list addressed, one fix commit per area so each stacked PR is a mechanical cherry-pick of its area commit plus its fix commit. The stack itself is the next step.

anishekkamal:
> The real-world score impact is much bigger than the PR body suggests, and it needs to be in the record.

Addressed: your nine-repo table is now in the PR description and in the CHANGELOG `Unreleased` entry under the v18 scoring change.

> I'll rule on the monotonicity-versus-responsiveness tradeoff separately ... hold off on recalibrating the harness until I confirm.

Acknowledged. Nothing in this round touches the calibration harness.

anishekkamal:
Split into the stacked PRs you asked for, in your order, each = its area commit plus its review-round fix commit (top of the stack is byte-identical to this branch after the fix round):

1. #119 output (drop `--format json` CodeDnaResult folded in)
2. #120 security lane
3. #121 cli/core (+ lint config, README enable/decline)
4. #122 session (resolve tombstones folded in)
5. #123 ml-client (beacon never rejects, report-open opt-out, init-guard scope-down)
6. #124 analyzers (Go unchecked-err rework, degraded marker, async overloads)
7. #125 codedna (single-pass tokenizer, dotted sinks, LSH star+chain, route tree freed)
8. #126 drift classification (raw-SQL anchor, logging hint scope, BASELINE_VERSION 7)
9. #127 scoring (v18, dedup key, mixed-group invariant, CHANGELOG + handbook; carries the calibration decision)

Every branch typechecks on its own and its area's tests pass. The starship Redundancy question stays open on this PR's `engine.ts` thread with my measurements. Closing this one in favour of the stack.

PR #119, opened 2026-09-01
fix(output): neutralize CSV formula injection, align the renderers, drop CodeDnaResult from JSON

anishekkamal:
**Stack 1 of 9.** Low risk, panel-verified.

- CSV export neutralises formula injection (`=`, `+`, `-`, `@` prefixes) and the renderers are aligned on the same finding shape.
- Review-round fix: `--format json` no longer dumps the raw `CodeDnaResult`, which embedded every extracted function's full source body. Drift findings and reimplementation candidates stay. Test asserts a function body never reaches the JSON.

Review items: the `--format json` P2. Validation on the full branch: typecheck and lint clean; 3049 tests pass, the 46 failures are the known Windows-only set (macOS/Linux green per review). Per-PR: typecheck clean and the area's test directories green (see the stack comment).

---
Stacked split of #113 (whole-repo audit), in the order requested in review. Each PR is its area commit from that branch plus its review-round fix commit; the top of the stack is byte-identical to `fix/whole-repo-audit`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_016KphwfjdmG6vFhxVye4dy7

PR #120, opened 2026-09-01
fix(drift): close false-bless and false-flag paths in the security lane

anishekkamal:
**Stack 2 of 9.** Bless/flag fixes verified against NEVER-FALSE-BLESS in review; both test inversions judged justified.

No review-round changes on this area. Validation on the full branch: typecheck and lint clean; 3049 tests pass, the 46 failures are the known Windows-only set (macOS/Linux green per review). Per-PR: typecheck clean and the area's test directories green (see the stack comment).

---

### DIFF
--- a/src/drift/route-extractors/go.ts
+++ b/src/drift/route-extractors/go.ts
@@ -40,7 +40,9 @@ function extractGoRoutesRegex(file: DriftFile, fileMiddleware: FileMiddleware | 
     if (!method || !path) continue;
 
     const context = lines.slice(Math.max(0, i - 10), i + 10).join("\n");
-    const handlerContent = findHandlerContent(file.content, path);
+    // Anchored to THIS route's line, so GET/POST on the same path and a path
+    // named in a doc comment above do not all read the first occurrence's window.
+    const handlerContent = findHandlerContent(file.content, path, i);
 
     const perAuth = GO_AUTH.test(context);
     const perVal = GO_VALIDATION.test(handlerContent);
--- a/src/drift/route-extractors/patterns.ts
+++ b/src/drift/route-extractors/patterns.ts
@@ -13,18 +13,42 @@
 export const GO_ROUTE_ECHO = /\.\b(GET|POST|PUT|PATCH|DELETE|HEAD|OPTIONS)\s*\(\s*"([^"]+)"/;
 /** Gorilla `HandleFunc("/x")…Methods("POST")` → capture [1] = path, [2] = METHOD. */
 export const GO_ROUTE_GORILLA = /HandleFunc\s*\(\s*"([^"]+)".*\.Methods\s*\(\s*"(\w+)"/;
-export const GO_AUTH = /[Aa]uth|[Tt]oken|require[A-Z]|middleware\.\w*[Aa]uth/;
+/** Go auth ENFORCEMENT shapes. Deliberately not the bare `[Aa]uth` / `[Tt]oken`
+ *  substrings this used to be: those matched `author`, `Authorization` header
+ *  WRITES, `oauthConfig`, and every `token, err := issueToken(...)` inside a login
+ *  handler, blessing the route that mints credentials as though it checked them.
+ *  Every alternate below names a middleware/guard, not a token-shaped noun. */
+export const GO_AUTH =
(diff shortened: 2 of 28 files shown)


anishekkamal:

skhan75:
Reviewed by running the branch, not by reading it: build, typecheck, lint (zero warnings), the full suite, `npm run calibrate` on main and on this branch, and real scans of 7 repos before and after (vibe-drift-api, vibedrift-landing-page, and five public Flask / FastAPI / Express / Koa / Gin apps). I also reverted each of the 11 fixes one at a time to check that its test goes red. All 11 do. Most of this PR is solid and lands as-is. Two things needed a change, and I pushed both to this branch so it can merge.

### 1. The "75 percent" gate change is not a bug fix, and it produced a wrong finding on a real repo (changed back)

**Plain version.** The detector only flags a route group when a strong majority of routes have a property. The code said "more than 75 percent". The handbook says the same in three places (`docs/handbook/06-security-consistency.md` lines 21, 175 and 320: "exceeds 0.75"). This PR changed it to "75 percent or more" and called that the documented gate. The only place ">= 75%" appears in the repo is the comment this PR added.

**Why it matters.** It changes the smallest group that can fire.

- Before: 4 of 5 routes (80%) fires, 3 of 4 (exactly 75%) stays quiet. Smallest group is 5.
- After: 3 of 4 fires. One stray route in a group of 4 is now a warning.

**What it did on vibe-drift-api.**

| | main | this PR as opened |
| --- | --- | --- |
| security_posture findings | 0 | 1: "Rate limiting missing on 1 of 4 routes" |
| route it cites | | `tests/test_plan_downgrade.py` (a test stub) |
| Security Posture | not measured | 19.7 / 20 |
| Vibe Drift Score | 75.0 | 82.1 |

All four routes in that group are stubs inside two test files. The "three rate-limited" ones are only rate limited because the test attaches `RequestSizeLimitMiddleware` (see point 2). So a false finding against a test fixture flipped a whole category into the score and moved it 7 points. The other 6 repos were byte-identical before and after.

**What I changed.** Put the gate back to `ratio <= 0.75` with a comment that explains why exactly 3 of 4 stays silent, and replaced the two boundary tests with one that pins the documented behaviour (3 of 4 silent, 4 of 5 fires; it fails if the gate is flipped to `<`). The two Python tests that were built on a 3-of-4 and 6-of-8 group got one more peer each so they clear the gate the documented way and still prove what they set out to prove. If you want 3-of-4 groups to fire, that is a product decision: it needs the handbook updated and a note that scores will move, not a silent change.

### 2. A bare "limit" still counts as rate limiting (fixed)

**Plain version.** `MIDDLEWARE_RATE_SEGMENTS` moved to whole-word matching, which correctly stops `UnlimitedAccessMiddleware` from counting. But it kept the word `limit` on its own. A limit is not a rate limit. I probed these names on the branch and every one returned `hasRateLimit: true`:

- `RequestSizeLimitMiddleware` (caps request body size)
- `BodyLimitMiddleware`
- `UploadLimitMiddleware`
- `ConnectionLimitMiddleware`

**What I changed.** Dropped `limit` and `limits` from the set (now `limiter`, `throttle`, `throttling`). Real rate limiters still bless: `RateLimitMiddleware`, `rate_limit_middleware` and `SlowAPIMiddleware` go through `RATE_NAMES`, which requires the `rate` prefix. Added a test with the four names above that fails if `limit` comes back, and extended the non-vacuity test.

### After both edits

- vibe-drift-api is back to 0 security findings and 75.0, the same as main. The other 6 repos are unchanged.
- Full suite green, typecheck and lint clean, `npm run calibrate` identical to main (security-floor gate 1.000).

### Follow-ups, not blockers for this PR

- **Routes inside test files count as production routes.** That is the other half of the false finding above. The security lane never applies `isAnalyzableSource`, unlike the other detectors. Worth its own issue.
- **`BASELINE_VERSION`.** The MCP baseline persists the security votes (`securitySubVotes` in `src/core/baseline.ts`), and this PR changes how those votes are computed. The bump is planned for #126. That is fine as long as nothing is released between this PR and #126; otherwise warm baselines keep the old votes.
- **`prunedDescendants`** was rewritten to survive very deep nesting, but no test feeds it deeply nested code, so the crash protection itself is unverified.
- A `/login` route can still be flagged as "missing auth". 0 of the 7 repos hit it, so it is a design note rather than a bug today.

### Verified sound

- File-level middleware now checks callee names, not raw argument text. `passport.authenticate()`, `jwt({ secret })` and `expressJwt()` still bless the file; `app.use("/jwt", jwtRouter)` no longer does.
- The Rust Actix bless reads the status from `HttpResponse::<Status>()` only; chained method names never contribute.
- The Python authn / authz split and the tighter Go / JS / Python auth regexes all move in the never-false-bless direction, and the regexes only run on the fallback path for files tree-sitter cannot parse.
- `findHandlerContent` has exactly one caller and it passes the line.
- The Go bucketing change is a pure refactor with the suite green.

---

skhan75 (approved):
Approved after the review-round edits above (gate restored to >75%, bare 'limit' no longer blesses rate limiting). Full suite, typecheck, lint and calibration green; all 7 real-repo scans identical to main. The release-rehearsal failure is the known verdaccio startup flake (also failing on main, fix in #133) and is not a required check.

---

PR #121, opened 2026-09-01
fix(cli): bind the report server to loopback, free WASM memory, make watch honest

anishekkamal:
**Stack 3 of 9.**

- Report server binds to loopback; parsed WASM trees are disposed after the last reader; `watch` reports what it actually re-scans.
- Also carries the lint-config/release-script changes and the README note on enable/decline from the audit branch's test/lint commit (the session-live deflake is in the session PR).

The route-auth-classify tree leak flagged in review is fixed in the codedna PR (stack 7), where that file's other changes live. Validation on the full branch: typecheck and lint clean; 3049 tests pass, the 46 failures are the known Windows-only set (macOS/Linux green per review). Per-PR: typecheck clean and the area's test directories green (see the stack comment).

---


anishekkamal:

skhan75:
Reviewed by running the branch: build, typecheck, lint (zero warnings), the full suite, real scans of 7 repos before and after, a memory measurement for the leak claim, and an inversion of each of the 10 fixes to check its test actually goes red. Most of this is good and lands as-is. One change had a real regression and three had no binding test; I pushed fixes for the ones worth fixing.

### 1. `--fail-on-score` stopped failing fast (fixed)

**Plain version.** `--fail-on-score` is the flag scripts use: it sets the exit code so a build can fail. Before this PR, when the score was below the threshold the process exited immediately with 1. This PR changed that to `process.exitCode = 1` and let the local report server own termination, so the command now sits there until someone presses Ctrl+C or the server's 10-minute timer fires.

The reasoning behind the change is right for a person at a terminal: the old code killed the report server about a second after starting it, so the report never loaded. The problem is that it applied to scripted runs too.

**Measured**, same tiny repo, `--fail-on-score 99` (score is 80, so it always fails), stdout piped:

| | exit code | elapsed |
| --- | --- | --- |
| main | 1 | ~0s |
| this PR | 1 | 40s and still running (timed out; would run the server's full 10 minutes) |

Same result with `CI=true`. Nothing in the serve path is gated on being interactive.

**Blast radius, measured.** Every documented path is safe, which is why this is a fix-and-land rather than a hold:

- README CI snippet: `--format terminal` → not the server path.
- `vibedrift hook install` pre-push hook: `--format terminal` → not the server path.
- The GitHub Action: `--json` → not the server path.
- A bare `vibedrift . --fail-on-score 80` (html is the default format, unauthenticated): **blocks**.

**What I changed.** Added `isInteractiveSession()` (no `CI` env var and stdout is a TTY) and kept the server alive only for that case. Non-interactive callers exit immediately, exactly as before. Re-measured: 1s and exit 1 both with and without `CI=true`. The existing test asserted only the new behaviour, so it passed either way; it now covers both, and dropping the gate makes the non-interactive case fail.

### 2. The path-traversal test did not bind (fixed)

`loadScanById` gained a guard rejecting a `--since` value with a separator or `..`, which is a good fix. But removing that guard leaves all four of its tests green.

The reason is arithmetic. The code builds `scan-${scanId}.json`, so the `scan-` prefix swallows the first `..`. The test's payload `../../secret/leak` resolves to `<home>/scans/<hash>/secret/leak.json`, while the test planted its file at `<home>/secret/leak.json`. The read fails either way, so the assertion passed for the wrong reason.

```
../../secret/leak         -> /HOME/scans/HASH/secret/leak.json
../../../../secret/leak   -> /HOME/secret/leak.json      <- actually reaches the planted file
```

**What I changed.** Moved the payload to four levels, with a comment recording why fewer levels prove nothing. Removing the guard now fails that test. (The `..`-alone and backslash cases stay as documentation; on POSIX neither can resolve to a real file, so neither can bind here.)

### 3. The `.vibedrift.json` weight override was removed but still documented (fixed)

The PR deletes `loadVibeDriftConfig` / `resolveDeviationWeights` because nothing ever called them, which is the right call — a documented setting that silently does nothing is worse than no setting. But two places still advertised it: `docs/handbook/07-code-dna.md` ("Weights are overridable per project via `.vibedrift.json`") and the header comment in `src/codedna/deviation-heuristics.ts`. The repo rule is same-PR handbook updates.

**What I changed.** Updated both to say the weights are fixed constants and why the loader was removed, then rebuilt the handbook with `npm run handbook`.

### Verified sound

- **Freeing the parsed trees causes zero behaviour change.** All 7 repos produce byte-identical JSON before and after (composite, findings, drift findings, per-file scores). Disposal is correctly ordered after the last reader: analyzers, then drift detection, then Code DNA, then dispose. I enumerated every `.tree` reader; `src/core/import-graph.ts` is the only one outside that group and it runs inside the analyzer and drift passes. Nothing downstream (scoring, rendering, history, the ML client, baseline assembly) touches `.tree`.
- **The memory claim is real but smaller than "unbounded".** Re-parsing the same 231 files five times in one process: RSS grew 973 MB on main and 457 MB with this PR, so about half. It is not eliminated, and peak RSS for a single scan is unchanged (4055 MB vs 4050 MB) since disposal happens at the end of the pipeline. The win is in repeated in-process scans (`watch`, the long-lived MCP server), which is what the comment describes.
- **8 of 10 fixes have a test that fails when the fix is reverted**: loopback bind, `semverGreater` prereleases (5 tests), the negated glob character class, the `-`-prefixed git ref, `disposeTrees`, the symlink-cycle guard, and both `--fail-on-score` branches after my change.
- Full suite 2866 passing, typecheck and lint clean, and lint now covers `scripts/*.mjs` (previously excluded entirely) at zero warnings.

### Not fixed, worth knowing

- **`parser.delete()` has no test.** Removing it leaves all 8 `ast.test.ts` tests green. Hard to unit-test a WASM free; the byte-identical scans and the memory measurement above are the evidence that it is both safe and effective.
- **The watch liveness probe is tested in isolation, not wired.** Replacing `probeRecursiveWatchLiveness(...)` with `true` at the call site keeps all 7 tests green, so the fallback-on-degraded-watch behaviour itself is unguarded.
- **The probe writes into the scanned repo.** It creates and removes `.vibedrift-watch-probe/` in the repo root at watch startup. It is in `IGNORE_SEGMENTS` and removed in a `finally`, but it will briefly appear in `git status` for anyone watching a clean tree.
- **`snapshotMtimes` now skips symlinks entirely**, so symlinked files are no longer watched in poll mode. That is the correct trade against the infinite-recursion cycle it fixes, just worth stating as a behaviour change.

---

skhan75 (approved):
Approved after the review-round edits above (--fail-on-score exits fast when non-interactive, traversal test now binds, handbook no longer documents the removed .vibedrift.json override). Full suite, typecheck, lint green; all 7 real-repo scans byte-identical to main. release-rehearsal is the known verdaccio startup flake, also failing on main (fix in #133), and is not a required check.

---

PR #122, opened 2026-09-01
fix(session): write sidecar state atomically and persist resolves

anishekkamal:
**Stack 4 of 9.**

- Sidecar state written atomically; the offline beacon hang is gone.
- Review-round fix (P1): resolves never persisted because `writeOutcomeState` merged `open` as a pure union with the on-disk copy. `OutcomeState` now carries a capped `resolved` tombstone list; the merge drops tombstoned ids from the disk copy and a re-flag clears its tombstone. Single-writer repro, re-flag, and legacy-sidecar tests added; driven end to end through the real hook.
- session-live test deflake.

---


anishekkamal:

skhan75:
Reviewed by running the branch: build, typecheck, lint (zero warnings), the full suite, real scans on two repos, a timing measurement of the offline-hang claim, and an inversion of each of the 9 fixes to check its test actually goes red. The changes are sound. Four of them had no binding test, including the wiring of the headline P1 fix; I pushed tests for three of them and one small test-only export to make the fourth testable.

### The offline hang is real, and the fix works

Measured on the same tiny repo with an unreachable API (`VIBEDRIFT_API_URL=http://127.0.0.1:9`), `--format terminal`, `CI=true`:

| | `--local-only` | beacon enabled |
| --- | --- | --- |
| main | 515 ms | **3443 ms** |
| this PR | 477 ms | **490 ms** |

The abort timer was only cleared on the success path, so a failed fetch (offline, DNS failure, connection refused) left a ref'd timer keeping the event loop alive for the full 3-second beacon timeout. Every scan on a machine with no network paid that. Moving `clearTimeout` into a `finally` removes it entirely.

Worth calling out separately because the PR body does not: **`sendReportOpenBeacon` now checks the telemetry opt-out**, which it never did before. That is a privacy fix, not just a cleanup — a user who ran `vibedrift telemetry disable` was still sending a report-open beacon. It has its own test and the test binds.

### The resolve-persistence fix was correct but its wiring was unguarded (test added)

The P1 this PR fixes: `writeOutcomeState` became a read-merge-write, and since a resolve is only an *absence* from `open`, a plain union read "local no longer has DF-1" as "local never heard of DF-1" and copied it straight back from disk on the very write meant to remove it. The tombstone list fixes that, and `mergeOutcomeState` is well tested — removing the tombstone filter fails 2 tests.

But the **wiring** was not. Deleting `outcomes.resolved.push(...resolvedIds)` from `hook-entry.ts` left all 436 session tests green, and so did deleting the re-flag tombstone clear. The existing integration test asserts a `resolve` **event** reaches the ledger, and that event is emitted either way. The sidecar is what the next hook run actually reads, and that is where the resurrection happens.

**What I added.** The integration test that drives the real hook now also reads the outcomes sidecar and asserts the resolved finding is gone from `open` and present in `resolved`. Deleting the tombstone push now fails it.

### Atomic write: the atomicity itself was untested (test added)

Replacing the tmp+rename in `writeFileAtomic` with a plain `writeFile` to the target left all 5 tests green — they only checked that the file ends up with the right contents, which an in-place truncate-then-write also achieves. The whole point of the helper is that the hook's 2-second self-timeout calls `process.exit(0)` without waiting for pending I/O, so an in-place write can be interrupted mid-truncate.

**What I added.** A test that asserts the destination's inode changes across two writes. A rename gives a new inode; an in-place write keeps it. That is the signature of the atomic replace, and it is deterministic. It also asserts no `.tmp` sibling is left behind.

### Cooldown merge: wiring untested (test added, plus a test-only export)

Replacing the read-merge-write in `writeState` with a blind overwrite left all 33 `check.test.ts` tests green. The existing concurrency test asserts only that the file stays parseable and usable, which a blind overwrite also satisfies.

Binding it needs the disk to change *between* a caller's read and its write, which no public entry point can interleave. I added `__test_writeCooldownState`, following the same `__test_` re-export pattern this repo already uses in `watch.ts` and `scan.ts`, and a test where a stale writer's state is written over a higher `nextFindingSeq` and a foreign cooldown key. With the merge it keeps both sides and never rewinds the sequence; with a blind overwrite it fails.

### Verified sound

- **Scan output is unchanged.** Both local repos produce byte-identical JSON to the pre-PR baseline.
- **Full suite 2903 passing**, typecheck and lint clean.
- **6 of 9 fixes already had a binding test**: the tombstone filter in `mergeOutcomeState` (2 tests), both beacon fixes, the trial-recap wall-clock deadline, and the intent-state read-merge-write. With my three additions that is 9 of 9.
- **The merge functions are conservative in the right direction.** `mergeIntentState` takes the max of `unrelatedEdits`, which under-counts when two processes each increment, but under-counting means fewer scope flags, never more. `mergeCooldownState` takes the max per key, so an earlier timestamp can never un-expire a throttle another writer started. `mergeOutcomeState` keeps a locally-open finding even when the disk copy says resolved, which matches the module's stated bias: a stale open finding costs a re-flag, a false resolve costs the user's trust.
- **The trial-recap deadline fails closed.** Running out of wall-clock returns `null` (no claims) rather than a partial sum presented as the whole story, consistent with the other skip paths in that walk.

### Not fixed, worth knowing

- **An interrupted write leaves a `.tmp` sibling behind.** If the process dies between the write and the rename, `<path>.tmp.<pid>.<n>` stays in the sessions directory forever, and nothing sweeps them. Consistent with `upload-state.ts`, which has the same shape, so this is pre-existing rather than new — but the session directory is the one place a killed hook is routine.
- **`MAX_RESOLVED = 200` tombstones per session.** Past that, the oldest tombstones drop and a finding could in principle be resurrected by a writer still holding it open. The comment reasons that ids are monotonic so the oldest stop mattering; not measured against a long real session.
- **Duplicate `DF-<n>` ids across concurrent writers are still possible** (two writers both read `nextFindingSeq: 5`, both mint DF-5, the merge takes max 6). Not introduced here — it is the documented collision in `decision.ts` — but the `max` merge does not close it either.

---

skhan75 (approved):
Approved after the review-round additions above: the resolve tombstone is now bound to the real hook path, writeFileAtomic's rename is bound by an inode assertion, and the cooldown read-merge-write has a deterministic test. Offline hang measured fixed (3443ms -> 490ms). Full suite 2903 green; scan output byte-identical. release-rehearsal is the known verdaccio startup flake, also failing on main (fix in #133), and is not a required check.

---

PR #123, opened 2026-09-01
fix(ml-client): stop uploading function bodies; harden auth URL and MCP inputs; beacon never rejects

anishekkamal:
**Stack 5 of 9.**

- Deep-scan uploads no longer include function bodies; API URL must be https; MCP inputs validated.
- Review-round fixes: (P1) the scan beacon awaited `resolveApiUrl()` outside its try, so a malformed `VIBEDRIFT_API_URL` killed every scan via an unhandled rejection; the whole beacon is now best-effort and can never reject. (P2) The dead `sendReportOpenBeacon` is removed and the real inline report-open beacon honours the telemetry opt-out at the one call site that passes `scanId`. (P2) The root-dir plausibility guard moved from the shared init core to the MCP adapter, so the interactive `vibedrift init` can initialise a marker-less new project while the MCP tool still refuses one.

---


anishekkamal:

skhan75:
Reviewed by running the branch: build, typecheck, lint (zero warnings), the full suite, real scans on two repos, a caller enumeration for the two security-shaped changes, and an inversion of each of the 8 fixes to check its test actually goes red. The work is sound and lands. Three fixes had no binding test, including the one that carries the actual privacy behaviour; I pushed tests for all three.

### Note on the merge with #122

This branch predates #122, and both touch `src/telemetry/beacon.ts`. I resolved it in this PR's favour, deliberately:

- #122 added a telemetry opt-out check to the exported `sendReportOpenBeacon`. That function has **zero production callers** (only tests referenced it) — the real beacon is an inline `<script>` in the generated HTML.
- This PR deletes the dead function and gates the *real* path instead, in `renderToFormat`, by withholding `scanId` when telemetry is off. Without a `scanId`, the beacon block is never emitted at all.

So the behaviour #122 intended is preserved and now actually reaches users. #122's `clearTimeout`-in-`finally` fix for `sendScanBeacon` survives untouched.

### 1. The privacy gate that ships had no test (test added)

Reverting the two-line gate in `scan.ts` to `const scanId = (result as ScanResultAugmented).__scanId` left **all 100** CLI and telemetry tests green. The only test that covered "report-open respects the opt-out" was #122's, and it tested the function this PR deletes.

**What I added.** A test that drives `logAndRender` down the authenticated `--output` path and asserts on what `renderHtmlReport` actually receives: with `VIBEDRIFT_TELEMETRY_DISABLED=1` no call carries a `scanId`, and with telemetry on at least one does (non-vacuity). Reverting the gate fails it.

### 2. `enable`'s root-dir guard was unbound (test added)

`assertPlausibleProjectRoot` has good unit tests of its own, but nothing tied it to the `enable` call site: deleting the call left all 17 tests in that area green. That matters because `enable` writes agent hook config into a fully caller-controlled `rootDir`.

**What I added.** A test that `run()` rejects a marker-less directory and writes nothing into it.

### 3. The fix-prompts timeout was unbound (test added)

Adding an `AbortController` with a 30-second timeout is the right fix — this call sits on the render path, so an API that accepts the connection and never answers held the scan open indefinitely. But removing either the `signal` or the `clearTimeout` left all 41 ml-client tests green.

**What I added.** A test asserting the signal reaches `fetch` un-aborted and that no timer is left pending afterwards, which is the same leak shape #122's beacon test guards. Removing either half fails it.

### Verified sound

- **The upload sanitizer change is correctly scoped.** `sanitizeResultForUpload` has exactly one caller (`scan.ts`, the dashboard log-scan upload). The deep-scan path does not go through it, so stripping `rawBody` / `declarationCode` / `bodyTokens` cannot starve the embedding call. The comment's distinction is real: bounded evidence snippets (`FileLocation.snippet`, drift `Evidence.code` capped at 3 lines per file, `TaintSink.expression` capped at 100 chars) are kept and pinned by an existing test; only the unbounded whole-body fields are dropped. Reverting the strip fails 3 tests.
- **The plaintext-URL refusal does not break anything real.** I searched `src/`, `test/`, `eval/` and `.github/` for a non-localhost `http://` API URL: the only hits are the tests that assert the rejection. `localhost`, `127.0.0.1` and `::1` stay allowed for local development. Reverting the guard fails 3 tests; reverting the stderr warning fails 1.
- **`patchConfig`'s promise-chain mutex binds** (a concurrent-patch lost-update test fails without it), and it is honest about its limit: it serialises writers inside one process only, and says so.
- **The `init` guard split is deliberate and correct.** The MCP adapter gates because its `rootDir` is agent-controlled; the core does not, so the interactive `vibedrift init` can still initialise a brand-new directory with no markers yet. Both sides are commented and the MCP side has 3 binding tests.
- **Scan output is unchanged** on both local repos, and the full suite is 2952 green with typecheck and lint clean.

### Not fixed, worth knowing

- **`assertSafeApiUrl` throws rather than falling back.** A malformed or plaintext `apiUrl` sitting in `~/.vibedrift/config.json` now makes every command that resolves an API URL throw, including ones that would otherwise work offline. That is the safe direction for a credential, but it is a hard failure on a config a user may not remember setting, and the message is the only guidance they get.
- **The stderr warning fires on every command** that uses a non-default API URL, including `--json` runs. It goes to stderr so it does not corrupt machine output, but a self-hosted user will see it on every invocation.

---

skhan75 (approved):
Approved after the review-round additions above (binding tests for the inline beacon opt-out, enable's root-dir guard, and the fix-prompts timeout). Beacon conflict with #122 resolved in this PR's favour: it deletes the dead sendReportOpenBeacon and gates the real inline path instead, so the opt-out now actually reaches users. Full suite 2952 green; scan output byte-identical. release-rehearsal is the known verdaccio flake, also failing on main (fix in #133), and is not a required check.

---

Issue #114, opened 2026-09-01
[bug] naming: one-word identifiers always classify as camelCase, misreading snake_case repos


skhan75:
## What happened

`detectConvention` (`src/analyzers/naming.ts:22-28`) tests conventions in a fixed order, and the camelCase pattern runs before snake_case:

```ts
if (/^[a-z][a-zA-Z0-9]*$/.test(name)) return "camelCase";
if (/^[a-z][a-z0-9_]*$/.test(name)) return "snake_case";
```

Any one-word lowercase identifier (`config`, `main`, `parse`, `run`) matches both patterns, so the order decides: every one-word name votes camelCase, unconditionally. In snake_case ecosystems (Rust, Python) most identifiers in a small file are single words, so whole files get classified as camelCase deviants.

## User-visible outcome

On `main` @ 9787cd3, `npx tsx src/cli/index.ts <starship-clone> --local-only --format json`:

```
naming | 238 files use camelCase while majority uses snake_case (H=0.61)
```

starship is a Rust repo. 238 of its 250 scanned files (95%) are flagged as deviating from the repo's own convention. Those files' "camelCase" identifiers are words like `config`.

## Magnitude, measured

Population: starship @ upstream HEAD (2026-09-01), 250 scanned files. Result: 238 of 250 flagged. PR #108 additionally measured that wiring `itemCount` into this analyzer (the remaining naming third of #104) would cost starship 1.3 composite points for this misclassification, which is why that PR deliberately excluded `naming` from volume scoring.

## Blast radius

This blocks the naming third of #104: volume amplification cannot ship for `naming` while one-word identifiers vote wrong. The same order bug affects every language, but it only becomes drift where snake_case is dominant (Rust/Python repos).

## Suggested direction

A one-word lowercase name is convention-ambiguous: it should vote for neither convention (return null or a distinct "ambiguous" bucket), so votes come only from names that actually discriminate (contain an underscore, or an internal capital).

## What was not verified

The 1.3-point figure is PR #108's measurement, not re-run here. Everything else above was executed on this machine.

Origin: measured during the #108 review (deliberate scoping-out of `naming`).

---
🤖 Filed by Claude (Claude Code) on behalf of @skhan75

PR #124, opened 2026-09-01
fix(analyzers): close language blind spots, surface analyzer crashes, rework Go unchecked-err

anishekkamal:
**Stack 6 of 9.**

- Language blind spots closed; analyzer crashes isolated.
- Review-round fixes: (P1) the Go unchecked-err rewrite measured +41% false flags; reworked so it adds no flag conditions over the original heuristic (same-line `if err := f(); err != nil` counts as checked; any `err` mention in the 5-line window counts as handled). (P1) A throwing analyzer now prints a stderr warning, is not counted as a cache miss, and the result carries `degradedAnalyzers` (also in `--format json`). (P2) The error-handling async body scan stops at a depth-0 `;` so overload/abstract signatures no longer walk into the next function. Security analyzer version 4; imports analyzer version 4 (rebase).

---


anishekkamal:

skhan75:
Reviewed by running the branch: build, typecheck, lint, the full suite, `npm run calibrate`, an inversion of each of the 6 fixes, and real scans of **9 repos** — three Go (gin, consul-template, a gin example app), three Python, two JS/TS, plus this org's own API and landing page. The Go rework and the crash isolation are good work. Two changes had measurable false-flag classes that the unit tests did not catch; I pushed fixes for both, with the numbers below.

### 1. The Go rework is a big improvement, but it did add one flag condition

The reduction is real and large. Unchecked-error flags, counted over every non-vendor `.go` file:

| repo | main | this PR |
| --- | --- | --- |
| gin | 39 | 18 |
| consul-template | 161 | 45 |
| gin realworld example | 21 | 3 |

But the description says "we intentionally add no new flag conditions", and that is not quite right. The original heuristic had `!nextLine.startsWith("return")`; the rework dropped it. That exclusion was covering Go's **named-return idiom**:

```go
func (r Render) Render(w io.Writer) (err error) {
    _, err = w.Write(bytesconv.StringToBytes(format))
    return
}
```

Assigning the named result and returning it *is* how the error propagates. Without the exclusion these flag. Measured: **12 lines across the corpus**, 9 of them in gin — half of gin's entire post-rework flag set (18).

**What I changed.** Restored the exclusion, but only for the **immediately following** line, which is exactly what the original single-line lookahead did. A `return` further down the 5-line window is a different statement, so a genuinely dropped error still flags:

```go
err := f()
fmt.Println("done")
return nil        // still flagged — correct
```

Both halves have a test, and each fails when its half is reverted. Re-measured: gin 18 → 9, consul-template 45 → 42, the example app unchanged at 3.

### 2. The naming change flipped Python repos to a camelCase "majority" (fixed)

Adding Python `function_definition` to the identifier extractor is the right idea — Python function names were invisible to the naming vote. But `detectConvention` tries the camelCase regex **before** snake_case, so any single all-lowercase word (`main`, `run`, `health`) classifies as camelCase. That is the open issue #114, and it was mostly harmless until this change fed it a large population of exactly that shape.

Measured on real Python repos, main vs this PR:

| repo | main | this PR | new finding |
| --- | --- | --- | --- |
| full-stack-fastapi-template | 85 | **84.8** | "25 files use snake_case while majority uses **camelCase**" |
| vibe-drift-api | 75 | 75 | "No dominant naming convention — snake_case (80 files), camelCase (46 files)" |
| flask-realworld-example-app | 87.2 | 87.2 | "7 files use camelCase while majority uses snake_case" |

The first one is backwards: it tells a Python project that camelCase is its convention and flags 25 files for using snake_case. `naming` is drift-kind, so it also moved that composite by -0.2.

**What I changed.** A single all-lowercase word is valid in *both* conventions, so it now votes for neither. That is the root cause of #114, and the naming analyzer's `version` is already bumped in this PR so warm caches invalidate.

Measured across all 9 repos with the fix in: **composite unchanged on 9 of 9** versus this PR. The findings get better:

- vibe-drift-api: "no dominant convention" → "**4 files use camelCase while majority uses snake_case**" (correct, and actionable)
- full-stack-fastapi-template: the backwards claim → "no dominant convention" (honest for a repo with a Python backend and a TS frontend)
- flask-realworld: the spurious finding goes silent
- the 4 Go/JS repos and the landing page: unchanged

### 3. `degradedAnalyzers` was unbound in the JSON (test added)

Crash isolation itself binds well — letting a throwing analyzer reject again fails a test. But removing `degradedAnalyzers` from `renderJsonOutput` left all 98 tests green, and that field is the only machine-readable signal that a score is missing an analyzer's findings. Added a test for both the present and absent cases.

### Verified sound

- **The error-handling body-brace fix recovers real coverage.** Async functions with an inline object return type (`Promise<{ ok: boolean }>`) were invisible to the old regex, which stopped at the type's `{`. The landing page has several in exactly the flagged directories, and its counts rise accordingly (12 → 26 and 23 → 32 in `src/lib/dashboard/` and `src/lib/admin/`). Reverting the fix fails 2 tests.
- **The naked-goroutine dominance gate works as designed**, and its practical effect is large: gin 20 → 0 and consul-template 41 → 0, because fewer than 60% of goroutines in either repo thread context, so there is no convention to deviate from. Worth knowing that on 2 of 2 Go repos with goroutines the detector now says nothing.
- **config-drift's bracket-notation patterns find real env vars.** `os.environ["OPENAI_API_KEY"]` in this org's API was previously invisible; it and two others now surface.
- **`getLineNumber`'s memoization is correct**, verified by the differential test against the naive implementation. It does not "bind" under inversion, which is expected for a pure performance refactor — the tests guard correctness, and the naive version is correct too.
- Full suite 2992 green, typecheck and lint clean, calibration identical to main with the security-floor gate at 1.000.

### Not fixed, worth knowing

- **The `degradedAnalyzers` assignment in `scan.ts` is still unbound.** Removing `result.degradedAnalyzers = pipeline.failedAnalyzers` leaves everything green; only the JSON serialisation half is now covered. Binding it needs a full-pipeline harness with an injected throwing analyzer.
- **The naming finding can report more files than were scanned.** vibe-drift-api showed "snake_case (80 files), camelCase (46 files)" against 104 files scanned, because a file with identifiers in both styles counts in both buckets. Pre-existing, but the new Python coverage makes it more visible.

---

skhan75 (approved):
Approved after the review-round edits above: an immediately-following return counts as handling a Go error again (12 named-return false flags across gin/consul-template), a one-word identifier votes for no naming convention (fixes the backwards camelCase-majority claim on Python repos, composite unchanged on 9 of 9 repos), and degradedAnalyzers is bound in the JSON. Full suite 2992 green; calibration identical to main. release-rehearsal is the known verdaccio flake, also failing on main (fix in #133), and is not a required check.

---

PR #125, opened 2026-09-01
fix(codedna): anchor taint regexes, correct the LCS gate, gate pattern drift on dominance

anishekkamal:
**Stack 7 of 9.**

- Taint sink/sanitizer anchoring; LCS length-ratio gate corrected; pattern drift gated on a dominance vote; interprocedural param names.
- Review-round fixes (P2s): strings and comments are recognised in a single left-to-right pass shared by both tokenizers (neither sequential order was correct: URL-in-string vs apostrophe-in-comment); bare sink anchors also accept known module/global receivers (`cp.exec`, `window.fetch`); an oversized LSH bucket emits star+chain pairs instead of dropping the whole cluster; `classifyRouteAuth` frees its parsed tree in a `finally` (the MCP-server leak flagged in review).

Note: this changes persisted MinHash signatures, so `BASELINE_VERSION` bumps in the drift PR (stack 8). Validation on the full branch: typecheck and lint clean; 3049 tests pass, the 46 failures are the known Windows-only set (macOS/Linux green per review). Per-PR: typecheck clean and the area's test directories green (see the stack comment).

---


anishekkamal:

skhan75:
Reviewed by running the branch: build, typecheck, lint, the full suite (3036 passing), `npm run calibrate`, scans of 9 real repos before and after with timings, and an inversion of each of the 7 fixes. Every one of the 7 has a test that fails when the fix is reverted, which is the best test-binding record of the stack so far. No changes needed from me. Two things belong in the record before this lands.

### The score moves on one repo, and the cause is the pattern-drift change

Scanned 9 repos against current `main`. Eight are byte-identical. One moves:

| repo | main | this PR |
| --- | --- | --- |
| vibe-drift-api | 75.0 | **73.3** |
| the other 8 | unchanged | unchanged |

The finding set is identical — same 115 findings, same 10 drift findings. What changed is one finding's *weight*. The `codedna-pattern` finding "api/models/embeddings.py uses raw_sql while 6 [others don't]" now carries a `driftSignal` (`consistencyScore: 86`, 6 of 7 files), so the scoring engine scores it as a dominance deviation instead of through the saturating-density branch. Its `consistencyImpact` goes 0.01 → 0.56 and Architectural Consistency drops 11.3 → 10.8.

That is the change working as designed — 6 of 7 files agreeing is a real dominance vote, and scoring it as one is the point. But it is a user-visible score movement coming out of a PR whose description reads as accuracy plumbing, so it should be stated rather than discovered on upgrade.

Timings on 4 repos are unchanged within noise (largest delta 127 ms on gin, ~5%), so widening the LCS gate has not cost anything measurable.

### The mixed-patterns hazard from the #113 review is still open

The earlier review flagged that `codedna-pattern`'s new `driftSignal` is defeated whenever the same scan also emits a signal-less finding under the same analyzer id. That mechanism is still present, and I confirmed both halves by reading the code:

- `pattern-classifier.ts:253` emits the pattern-drift finding **with** a `driftSignal`.
- `pattern-classifier.ts:272` emits "Mixed patterns in …" under the **same** `analyzerId: "codedna-pattern"` with **no** `driftSignal`.
- `engine.ts:485` gates the dominance branch on `findings.every((f) => f.driftSignal)`.

So a single mixed-patterns finding anywhere in the scan flips the whole group back to the density branch and silently cancels this PR's scoring fix.

**Magnitude: not observed.** Across all 9 repos, only one produced any `codedna-pattern` finding at all (vibe-drift-api: 1 pattern-drift, 0 mixed), and the dominance branch engaged there. I also built a synthetic fixture aiming to trigger both finding kinds in one scan and could not get the classifier to emit either, so I could not construct the failing input. **UNVERIFIED** whether this is reachable on real code. The PR body says the mixed-group invariant is pinned in the scoring PR (stack 9); pinning it there tests the engine's behaviour but does not stop the interaction, so it is worth deciding explicitly whether the mixed finding should get its own analyzer id.

### Verified sound

- **The LCS gate correction is right, and the old comment was wrong.** The old gate rejected at `r < 0.5` claiming similarity "cannot exceed 0.5" there, but the true ceiling for a length ratio `r` is `2r/(1+r)`, which is 0.667 at r = 0.5. Everything from r = 1/3 up to r = 0.5 was returned as 0 despite a real similarity up to 0.645 — above the 0.60 discovery threshold and the sampler's 0.55 band floor, which is exactly the wrapper-and-inline duplication shape. The new default `minSimilarity` of 0.5 sits at or below every caller in the tree (0.55 sampler, 0.60 discovery, 0.70 flagging, 0.80 validation), so no caller loses a pair it would have flagged. Reverting fails 2 tests.
- **The single-pass tokenizer is the only correct shape**, and the comment explains why both sequential orders are wrong. Reverting to the sequential strip fails 2 tests, one for a URL inside a string literal and one for a block comment containing `//` and a quote.
- **The oversized-LSH-bucket handling now emits star+chain instead of dropping the cluster.** Reverting to a plain `continue` fails the test that asserts O(m) connectivity rather than nothing.
- **The permutation-seed table grows on demand.** It was fixed at 128 and indexed `p % 128`, so asking for more than 128 permutations silently produced duplicate, perfectly-correlated hash rows while the LSH math assumed independence. Reverting fails a test.
- **Taint sink anchoring accepts real dotted receivers** (`cp.exec`, `child_process.exec`, `window.fetch`) without going back to bare substring matching. Removing the receiver list fails 3 tests.
- **`classifyRouteAuth` frees its parsed tree in a `finally`** — the MCP-server leak flagged in review, and the one place `parseFile` hands back a tree nobody else owns. Making the `finally` a no-op fails its test.
- **Calibration is identical to main**, security-floor precision gate at 1.000.

### Merge-order note

This changes persisted MinHash signatures, and `BASELINE_VERSION` bumps in the drift PR (stack 8). Nothing should be released between this PR and that one, or warm MCP baselines will keep signatures built by the old tokenizer.

---

skhan75 (approved):
Approved as-is: all 7 fixes have a test that fails when the fix is reverted, the best binding record of the stack. Recorded in the review above: the pattern-drift dominance signal moves vibe-drift-api 75.0 -> 73.3 (8 of 9 repos unchanged), the mixed-patterns hazard from the #113 review is still open with magnitude unverified, and BASELINE_VERSION must land in #126 before any release. release-rehearsal is the known verdaccio flake, also failing on main (fix in #133), and is not a required check.

---

PR #126, opened 2026-09-01
fix(drift): rank classifications by evidence, fail hint seeding closed, BASELINE_VERSION 7

anishekkamal:
**Stack 8 of 9.**

- Classifications ranked by evidence; intent-hint seeding fails closed.
- Review-round fixes: (P1) the raw-SQL regex matched import lines and prose (this repo's own `html.ts`/`terminal.ts` classified as raw SQL); each statement form must now open a string literal and name an identifier, with tests that read those two files from disk. (P1) A language-exclusive logging hint (`go_slog`) no longer seeds the vote on other languages' partitions. (P2) `convention-oscillation` uses a symbol count for both numerator and denominator again. `BASELINE_VERSION` 7 (tokenizer change alters persisted signatures; vote shapes gained `allDominantFiles`).

---


anishekkamal:

skhan75:
Reviewed by running the branch: build, typecheck, lint, the full suite (3091 passing), `npm run calibrate`, scans of 6 real repos before and after against current `main`, and an inversion of each substantive fix. All six bind. No changes needed from me. The one thing that must be in the record is that **this PR raises scores on 4 of 6 repos**, and why.

### Measured score impact

| repo | main | this PR | change |
| --- | --- | --- | --- |
| vibe-drift-api | 75.0 | 78.2 | **+3.2** |
| full-stack-fastapi-template | 84.8 | 87.4 | **+2.6** |
| node-express-realworld-example-app | 88.2 | 89.2 | +1.0 |
| vibedrift-landing-page | 90.2 | 90.9 | +0.7 |
| gin | 97.0 | 97.0 | 0 |
| consul-template | 84.0 | 84.0 | 0 |

Every movement is upward, and every one traces to a **false finding being removed** rather than a threshold being loosened. The two largest:

**Cross-axis intent seeding.** A single `architectural_consistency` hint in CLAUDE.md was seeding all four axes — data access, error handling, config, and DI. A "repository pattern" declaration says nothing about error handling, but it injected a phantom entry into that vote, and because a seeded vote skips the dominance threshold, the axis emitted a finding the raw vote would never have produced. On vibe-drift-api that was "3 file(s) use error swallowing while 11 use exception throwing", which disappears here. Binding the seed to the axis whose pattern record actually holds the key is the right fix, and reverting it fails a test.

**Logging partitioned by language.** Three of the five logger families are language-exclusive, so a pooled vote made every TypeScript file in a Python-majority repo "deviate from the Python logging module". On vibe-drift-api the finding goes from "12 file(s) use Python logging module, console.* while 23 use a shared structured logger" to "7 file(s) use Python logging module while 23 use a shared structured logger" — the TypeScript files drop out, correctly. On full-stack-fastapi-template the logging finding disappears entirely.

The related hint fix is the same shape: a repo-wide `go_slog` declaration used to seed every partition, so TypeScript files were told "team declared Go log/slog but 3/3 files use console.*" with a "Migrate 0 file(s)" recommendation. Now a hint may only seed partitions whose language could have honoured it.

### Verified sound

- **The raw-SQL classifier was badly over-matching and now is not.** The old pattern `(?:SELECT|INSERT|UPDATE|DELETE)\s+(?:FROM|INTO|SET|\*)\b` matched none of `SELECT * FROM users`, `SELECT id FROM users`, or `UPDATE users SET x = 1` — while a bare keyword-pair search over whole files bridged `import { x } from "./select.js"` to the next `from` and classified prose comments. Anchoring each statement form on a string-literal quote, with per-statement case consistency, fixes both directions. Reverting fails **13** tests.
- **The return-shape vocabulary guard is real.** `seedDominanceVote` injects a declared pattern the distribution does not hold as a zero-count entry weighted ~1.95, which outvotes a single real file — so an out-of-vocabulary declaration won the vote with a count of zero, drove `consistencyScore` to 0, and rendered an undefined shape name. Reverting fails a test.
- **The convention-oscillation ratio no longer mixes units.** The numerator counted symbols and the denominator counted files, so the rendered "X of Y" could exceed 100%. Both are symbol counts now, and the comment is honest that this makes the scoring engine's sample-confidence term trivially satisfied for symbol axes — a pre-existing tradeoff it explicitly declines to fix here. Reverting fails 3 tests.
- **`BASELINE_VERSION` is bumped to 7**, which this stack needs: #125 changed persisted MinHash signatures and this PR changes vote logic. The version *mechanism* is well covered — one test asserts `assembleBaseline` stamps the current version, another asserts the MCP provider rebuilds once when an on-disk baseline predates it. Both reference the constant symbolically, so neither pins the literal 7; setting it back to 6 leaves all 165 core tests green. That is correct for those tests, but it means nothing mechanical stops a rebase from dropping the bump. Worth a second look at merge time rather than a tautological test.
- Calibration identical to main, security-floor precision gate at 1.000.

### Merge-order note

With this merged, the stack's baseline gate is in place: #125's signature change and this PR's vote changes are both covered by `BASELINE_VERSION = 7`. A release may now be cut from `main` without stranding warm MCP baselines.

---

skhan75 (approved):
Approved as-is: all six substantive fixes have a test that fails when the fix is reverted. Recorded in the review above: this PR raises scores on 4 of 6 repos (+0.7 to +3.2), every movement traced to a false finding removed rather than a threshold loosened, and BASELINE_VERSION 7 now covers #125's signature change plus this PR's vote changes. release-rehearsal is the known verdaccio flake, also failing on main (fix in #133), and is not a required check.

---

PR #127, opened 2026-09-01
fix(scoring): make the composite monotonic in findings (SCORING_VERSION v18)

anishekkamal:
**Stack 9 of 9. Reshapes every user's score.**

- The composite is provably non-increasing in the finding set: surface-specific categories enter clamped to the always-measured geometric mean; a category with findings is capped at its own empty-category credit; detector confidence aggregates as the group max.
- Review-round fixes: (P1) ML duplicate findings carry no line numbers, so the file:line dedup key never merged them with their Code DNA twin; a line-less finding now merges by file set into the lined group over those files. (P1) A detector group mixing `driftSignal` and signal-less findings (reachable via `codedna-pattern`) now scores as the max of its dominance and count parts, with the sample-size discount folded into the dominance term; a 200-trial randomized property test pins that removing any finding never lowers the composite.
- `SCORING_VERSION` v18 with a history entry, CHANGELOG `Unreleased`, handbook ch. 07/08 refreshed.
- **Merge-round addition:** `calibrate:monotonic` now gates on monotonicity alone and reports responsiveness without enforcing it. See below.

## Real-world score impact (re-measured against CURRENT main, 13 repos)

The original table in this description used pre-stack `main` as its base. Eight PRs have merged since, several of which moved scores on their own, so those numbers no longer describe this PR. Re-measured against `main` as it stands today, which is what an upgrading user actually experiences:

| repo | main today | with this PR | change |
| --- | --- | --- | --- |
| moment | 68.0 | 56.3 | -11.7 |
| date-fns | 67.2 | 56.7 | -10.5 |
| zod | 75.7 | 67.4 | -8.3 |
| htmx | 76.4 | 68.4 | -8.0 |
| vibedrift-landing-page | 90.2 | 86.7 | -3.5 |
| gin | 97.0 | 95.5 | -1.5 |
| full-stack-fastapi-template | 84.8 | 83.8 | -1.0 |
| dayjs | 80.9 | 80.9 | 0 |
| underscore | 54.4 | 54.4 | 0 |
| consul-template | 84.0 | 84.0 | 0 |
| vibedrift-public | 79.9 | 80.2 | +0.3 |
| starship | 96.2 | 97.2 | +1.0 |
| vibe-drift-api | 75.0 | 77.9 | +2.9 |

**10 of 13 move, 7 down and 3 up, from -11.7 to +2.9.** The earlier table's headline starship jump of +25.5 was the stack's, not this PR's: main already has starship at 96.2 and this adds 1.0. The `itemCount` question left open on #113's `engine.ts` thread therefore does not gate this merge.

**Five repos drop a letter grade**: zod and htmx C to D, moment and date-fns D to F, the landing page A to B. vibedrift-public gains one, C to B.

For 10 of the 13 repos the new composite is simply the geometric mean of the categories that actually found something. moment is the clearest case: Redundancy 7.2/20, Architectural Consistency 17.7/20, Intent Clarity 19.7/20 because that detector had little to report. The old formula averaged all three to 68; the new one refuses to let a near-empty category lift the headline above what the measured categories say, and lands on 56. The two repos that rise do so for a different reason — the dedup fix removes findings that two layers were counting twice, which lifts their Architectural Consistency.

Note the asymmetry this creates: Intent Clarity and Security Consistency can now lower a score or leave it alone, but never raise it. A repo with genuinely excellent, consistent auth gets no credit for it in the headline.

## The calibration harness, split

`calibrate:monotonic` was checking two unrelated things behind one exit code, which is why it sat red for months with neither getting fixed:

| | main today | this PR |
| --- | --- | --- |
| monotonicity | ✗ 0% (86.7) → 10% (87.4) | ✓ (gates) |
| responsiveness | ✓ | ! 1.4 at 25→50, target ≥3 (report-only) |
| exit code | 1 | **0** |

Monotonicity is a correctness property and now gates alone. Responsiveness is a sensitivity target: its 3-point threshold was calibrated against pre-v18 scoring, and re-deriving it against today's fixture would be fitting the gate to the test, so it is reported and tracked rather than enforced. That is a deliberate call to be revisited with the scoring-algorithm work, not a silent loosening.

Every run now prints the monotonicity margin. It is currently **0.0**, and instrumenting the fixture shows why that is benign: injecting drift grows the corpus from 369 to 374 lines, which lifts every category's evidence-weighted clean credit by 0.005, and `securityPosture` enters at exactly the base mean and contributes nothing.

```
0%   arch 0.825  redundancy 0.825  security N/A    -> composite 82.5
10%  arch 0.830  redundancy 0.830  security 0.830  -> composite 83.0
```

So the +0.5 step is the fixture getting bigger, not the score rewarding drift. The fixed-corpus property this PR actually claims is pinned separately by `monotonicity.test.ts` across 200 randomized trials, and every one of its assertions fails when the corresponding fix is reverted.

---
Stacked split of #113 (whole-repo audit). This is the last of the nine.


anishekkamal:

skhan75:
Reviewed by running the branch on top of current `main` (all eight earlier stack PRs merged): build, typecheck, lint, the full suite (3107 passing), both calibration harnesses, an inversion of each of the five substantive fixes, and scans of **13 real repos** including every repo in this PR's own table. Every fix binds. The code is correct and the reasoning in the v18 history entry is sound.

**Not approving to merge yet**, because two things in the description no longer match what I measure, and the product decision should be made against the real numbers rather than these.

### 1. The score table is stale — it measures the whole stack, not this PR

The table's base is pre-stack `main`. Eight PRs have merged since, and several of them moved scores on their own. Re-measured against **current main**, which is what a user upgrading will actually experience:

| repo | this PR's table | main today | with this PR | actual change |
| moment | 68.2 → 56.6 | 68.0 | 56.3 | **-11.7** |
| date-fns | 67.0 → 56.7 | 67.2 | 56.7 | **-10.5** |
| zod | 75.7 → 67.4 | 75.7 | 67.4 | **-8.3** |
| htmx | 76.4 → 68.4 | 76.4 | 68.4 | **-8.0** |
| **starship** | **71.7 → 97.2 (+25.5)** | **96.2** | **97.2** | **+1.0** |
| dayjs | 80.6 → 80.6 | 80.9 | 80.9 | 0 |
| underscore | 54.5 → 54.5 | 54.4 | 54.4 | 0 |
| vibedrift-landing-page | not listed | 90.2 | 86.7 | -3.5 |
| gin | not listed | 97.0 | 95.5 | -1.5 |
| full-stack-fastapi-template | not listed | 84.8 | 83.8 | -1.0 |
| vibe-drift-api | not listed | 75.0 | 77.9 | **+2.9** |
| vibedrift-public | 86.2 → 81.5 | 79.9 | 80.2 | +0.3 |
| consul-template | not listed | 84.0 | 84.0 | 0 |

Six of the seven listed numbers reproduce almost exactly, which is good. The one that does not is the headline: **the starship jump is not this PR's**. By the time this lands, main already has starship at 96.2; this PR adds +1.0. The open question on #113's `engine.ts` thread about whether count-based duplicate findings should carry `itemCount` therefore no longer gates this merge.

Summary against current main: **10 of 13 repos move, 7 down and 3 up, from -11.7 to +2.9.**

The v18 history entry states "the direction is always down or unchanged". That holds for the monotonicity repair in isolation, but not for this PR as a whole: three repos rise, up to +2.9, because the dedup and mixed-group fixes also move scores. Worth correcting in the comment so the next reader is not misled.

### 2. Five of 13 repos drop a letter grade

This is the part users will actually notice:

| repo | grade change |
| --- | --- |
| zod | C → **D** |
| htmx | C → **D** |
| moment | D → **F** |
| date-fns | D → **F** |
| vibedrift-landing-page | A → **B** |
| vibedrift-public | C → B (up) |

The mechanism on the biggest mover is worth stating plainly. On moment the category scores barely change — Architectural Consistency 17.7 → 17.6, Intent Clarity 19.7 → 20.0, which goes *up*. The composite still falls 11.7 points because Intent Clarity's **contribution** is now clamped to the geometric mean of the always-measured categories. A repo whose surface-specific category sat well above its mean was being lifted by it, and that lift is exactly the inversion this PR removes. So the drop is the fix working, not a regression — but "your Intent Clarity went up and your grade went down" is a hard thing to explain, and the release notes will need to.

### 3. The monotonicity harness still exits 1, for the opposite reason

Ran `npm run calibrate:monotonic` on both trees:

| | main today | with this PR |
| --- | --- | --- |
| monotonicity | ✗ 0% (86.7) → 10% (87.4) | ✓ |
| responsiveness | ✓ | ✗ 25→50 drop is 1.4, needs ≥3 |
| exit code | 1 | 1 |

So the PR does fix the long-standing monotonicity failure, and the command stays red because the other gate now fails. That is the tradeoff you flagged you were leaning toward accepting; it is now measured rather than projected.

One caveat on the pass: the harness tolerates a rise of up to +0.5 (`cur.composite > prev.composite + 0.5`), and this branch's 0% → 10% step is 82.5 → 83.0, exactly +0.5. It passes at the boundary. That is not a contradiction of the "provably non-increasing" claim — the harness injects drift by rewriting files, which changes LOC and category applicability, so it is not the fixed-corpus add-a-finding comparison the property test pins — but the claim and this table are not measuring the same thing, and it is worth knowing the margin is zero.

### Verified sound

Every substantive fix has a test that fails when the fix is reverted:

- **Group confidence as max, not mean.** With a mean, adding a low-confidence finding *lowered* the group's damage, so a new finding improved the score. Reverting fails 3 monotonicity tests.
- **The surface-specific clamp.** Reverting fails 2 tests, including the named `65.0 → 74.9` inversion and the one asserting the Fix Plan never projects a lower score than doing nothing.
- **A category with findings capped at its own empty-category credit.** Reverting fails the test that a category holding findings never out-scores the same category holding none.
- **The mixed-group max.** This closes the hazard I flagged on #125: `codedna-pattern` emits both a signalled pattern-drift finding and a signal-less "Mixed patterns" finding under one analyzer id, and the old `every(driftSignal)` switch let one signal-less finding flip the whole group. Reverting fails a randomized property test.
- **The dedup fix for line-less ML findings.** Reverting fails 3 tests, including the one asserting a line-less finding attaches to exactly one lined group over the same files.
- `SCORING_VERSION` is v18 with a full history entry, CHANGELOG has an Unreleased section, and handbook chapters 07 and 08 are refreshed and rebuilt. Calibration precision is identical to main with the security-floor gate at 1.000.

### What I need from you

1. Accept the monotonicity-over-responsiveness tradeoff, on these numbers rather than the projected ones. If yes, the responsiveness thresholds in `run.ts` need recalibrating in a follow-up, otherwise the command stays red and stops signalling anything.
2. Accept 5 of 13 repos dropping a letter grade, and confirm the release notes will carry it. The one-time scoring-refined notice fires on version change, but it does not explain a grade drop.
3. Confirm the description should be updated to the current-main numbers before merge, so the permanent record does not credit this PR with starship's +25.5.

Happy to merge as soon as you rule on 1 and 2.

---

skhan75 (approved):
Approved. All five scoring fixes have a test that fails when the fix is reverted, and the description now carries the numbers re-measured against current main rather than pre-stack main.

Merging with the calibration harness split, per the maintainer's call: monotonicity gates alone (green), responsiveness is reported and tracked but not enforced until the scoring-algorithm work sets its threshold deliberately. The command exits 0 for the first time in months. The 0.0 monotonicity margin is instrumented and benign — the harness's fixture grows 369 to 374 lines when drift is injected, which lifts every category's evidence-weighted clean credit; securityPosture enters at exactly the base mean and contributes nothing.

Accepted with this merge: 5 of 13 repos drop a letter grade, and Intent Clarity / Security Consistency can no longer raise a score. Both need to be in the release notes.

---

